PR #70, opened 2024-05-03
rename `append` to `extend`

### LINKED ISSUE #64: Naming: Consider changing `WorkGraph.append`
The `WorkGraph.append` method allows adding another existing `WorkGraph` to it. The term can be confusing as `append` suggests order and one workgraph is subordinate to the order. However, the order is _only_ added later when the links are established. The order for the append method is arbitrary. For example:
```python
wg1 = WorkGraph()
wg2 = WorkGraph()
wg1.append(wg2)
wg1.links.new(...)
```
is identical to 
```python
wg1 = WorkGraph()
wg2 = WorkGraph()
wg2.append(wg1)
wg1.links.new(...)
```
i.e. it is arbitrary whether `wg1` is appended to `wg2` or vice version. Really, we are creating a "union" of two existing graphs, merging the nodes of one graph to the other. In a sense, the workgraphs are sets of nodes, and `WorkGraph.append` is adding the nodes of one workgraph to the other.

Suggested alternatives for the method are:
* `add`
* `union`
* `merge`

Another suggestion was to, instead of having the operation as a `WorkGraph` method, to have a function that merges multiple workgraphs into a new one. Something like
```python
def combine_work_graphs(*args: WorkGraph) -> WorkGraph:
     """Combine the nodes of multiple workgraphs into a new workgraph."""
```

superstar54:
WorkGraph has nodes and links collections, which are like lists. When combining WorkGraph B to WorkGraph A, it basically adds all the nodes (links) from B to the nodes (links) of A. 

In Python's list, `append` adds a single element to the end of the list, while `extend` adds multiple elements by appending each element of an iterable individually. Thus `extend` is more suitable in this case.

### DIFF
--- a/aiida_workgraph/workgraph.py
+++ b/aiida_workgraph/workgraph.py
@@ -337,8 +337,8 @@ class WorkGraph(node_graph.NodeGraph):
         self.ctx = {}
         self.state = "CREATED"
 
-    def append(self, wg, prefix=""):
-        """Append a workgraph to the current workgraph.
+    def extend(self, wg, prefix=""):
+        """Add a workgraph to the current workgraph.
         prefix is used to add a prefix to the node names.
         """
         for node in wg.nodes:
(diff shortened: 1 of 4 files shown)


agoscinski:
Do you have an example? I tried look through the repo but could not find usage. It sounds for me like one should support both. `append` for single nodes and `extend` for lists of nodes / graphs

PR #75, opened 2024-05-11
add `add_node` and `add_link`

### LINKED ISSUE #60: wg.add_node()
move methods (new, delete, etc) from nodes collection to workgraph.

superstar54:
- Add `add_node` and `add_link`
- Make them the default in the docs

Note: the original `nodes.new` and `links.new` is still kept.


aiida_workgraph/workgraph.py
```diff
@@ -371,3 +372,13 @@ def _repr_mimebundle_(self, *args, **kwargs):
             return self._widget._repr_mimebundle_(*args, **kwargs)
         else:
             return self._widget._ipython_display_(*args, **kwargs)
+
+    def add_node(self, identifier: Union[str, callable], name: str, **kwargs) -> None:
```
edan-bainglass: `add_task`?

edan-bainglass (changes requested):
Update `add_node` method to `add_task` following new convention.

## A bit more context

It is important that these new AiiDA tools keep the end user not only in mind but of utmost importance. Interface like `add_task` and `add_link` don't add much to the code. However, they can have a serious impact on user experience. `add_task()` (one method) vs. `tasks.new()` (two methods) is actually a big difference w.r.t UX.

#### NOTE: Highly opinionated 😎

aiida_workgraph/workgraph.py
superstar54: Yes, thanks!

edan-bainglass (approved):
LGTM!

PR #76, opened 2024-05-13
replace node with worknode

### LINKED ISSUE #61: Naming: Consider renaming `WorkGraph.nodes`
The `WorkGraph.nodes` property returns a collection of "nodes" of the workgraph. These nodes are _not_ the same as nodes in AiiDA's provenance graph and this is bound to cause confusion. Nodes in the workgraph context are really the subprocesses or steps of the workgraph. In the provenance graph, nodes can also refer to the inputs and outputs of the processes.

The suggestion is therefore to find another name for the nodes property of `WorkGraph`. Since conceptually they refer to the steps of the workgraph (or workflow) it might make sense to use a word in that direction:

* `WorkGraph.processes`
* `WorkGraph.steps`
* `WorkGraph.operations`

superstar54:
## WorkNode
`WorkNode` is the basic building block of the WorkGraph. A `WorkNode` has inputs, outputs, and an executor. Here is how the user defines and uses `WorkNode`:

```python
from aiida_workgraph import worknode, WorkGraph

# define a normal worknode
@worknode()
def add(x, y):
    return x + y

# define a calcfunction worknode
@worknode.calcfunction()
def multiply(x, y):
    return x*y

# visualize the worknode in Jupiter-notebook
add.worknode()

# Create a workgraph to link the nodes.
wg = WorkGraph("test_add_multiply")
wg.nodes.new(add, name="add1")
wg.nodes.new(multiply, name="multiply1")
wg.links.new(wg.nodes["add1"].outputs["result"], wg.nodes["multiply1"].inputs["x"])
```
## Compare with other AiiDA Node
- `WorkNode` is `node` in the `WorkGraph`.
- AiiDA node is `node` in the database. Even though the `Node` concept is complex in the API, for example, Data Node, ProcessNode, WorkChainNode, and differences between Process and its Node.

To distinguish from AiiDA Node, we always use `Work` and `Node` together (WorkNode).


superstar54:
Replaced by #108

PR #121, opened 2024-06-26
♻️ Rename `ShellTask` to `ShellJob`

### LINKED ISSUE #118: Revert naming of `ShellJob` to `ShellTask`
Recently, the name of the identifier for running shell commands via `aiida-shell` in `WorkGraph` has been renamed from `ShellJob` to `ShellTask`. While I understand where the motivation comes from (as part of the general rename from `Node` to `Task`), I don't think this is a good idea. It introduces inconsistencies, e.g. with `verdi process list`, as the actual entities that are submitted are still `ShellJob`s. This class name is defined inside `aiida-shell` and likely won't change. Furthemore, I don't think there is any need for the rename from `ShellJob` to `ShellTask` (the previous name didn't contain `Node` anyway).

GeigerJ2:
The previous renaming was done during the refactor from `Node` to `Task`, however, `ShellJob` is the name of the corresponding class in the `aiida-shell` plugin and here to stay. So the current naming will lead to inconsistencies, which is why this commit reverts it.

Fixes #118.


superstar54 (approved):
Great, thanks!

PR #164, opened 2024-07-15
Support for adding inputs/outputs/arguments after ShellJob initialization

agoscinski:
The ShellJob requires the inputs, arguments and outputs to be set on initialization. With the change on preparation we can add these later on to the workgraph. On preparation the added nodes to workgraph are retrieved and forwarded to the ShellJob.

The example was not properly run, because I cannot submit jobs for some reason (I get 302 and 303 Files are not found errors when I do this), running them works fine. Also checked with release version, so I assume it is an issue with the daemon or rabbitmq. Maybe @superstar54 you can run the notebook and check if works on your setup?

EDIT: Is there version control friendly way to commit the notebooks in this repo? Or do you just clear all output?

### DIFF
--- a/aiida_workgraph/engine/utils.py
+++ b/aiida_workgraph/engine/utils.py
@@ -131,12 +131,33 @@ def prepare_for_shell_task(task: dict, kwargs: dict) -> dict:
         if key.startswith("nodes."):
             nodes[key[6:]] = kwargs.pop(key)
     metadata.update({"call_link_label": task["name"]})
+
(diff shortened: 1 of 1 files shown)


agoscinski:
This PR works, but the implementation is a bit hacky. Since on preparation we cannot differ between parser_outputs and regular ShellJob outputs, we make it dependent if a parser was given. When mixing parser and ShellJob outputs however this will fail. Since a proper implementation will take considerable amount of work with not much additional benefit we made an issue #168 and close this PR.

PR #165, opened 2024-07-15
Make widget feature optional

agoscinski:
At the moment the widget part of workgraph is compulsory to install which requires to build JavaScript code. This commit makes the use of the widget optional and thus facilitates the installation of a development environment.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,7 +25,7 @@ jobs:
                 python-version: '3.10'
 
         -   name: Install Python dependencies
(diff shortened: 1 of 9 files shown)


superstar54:
Hi @agoscinski , thanks for the work. 

Could you please also update the `Installation` section in the docs, that if the user want to use the widget
```
pip install aiida_workgraph[widget]
```

agoscinski:
Updated

superstar54 (approved):
I made changes to handle _repr_mimebundle_ when `_widget` is None

PR #186, opened 2024-07-18
Add executed_task to record the tasks that have already been submitted. 

superstar54:
## Issue
@edan-bainglass reported an error in his testing: Some of the tasks were run multiple times. I can confirm that this is a bug.

## Reason
In WorkGraph engine, `calcfunction` is run directly instead of submit. In order to not block the child task of a `calcfunction` task, the engine will continue to run the workgraph (`self.continue_workgraph` method) immediately after the execution of a `calcfunction`, until all its child tasks are finished, or going to a `Waiting` state. This will run recursively.  When continuing the workgraph, we pass a `exclude` list to avoid running the tasks that will be executed in the current run. However, the `exclude` list only considers the current run instead of the recursive run.

More generally, there is a risk that the engine can run a task multiple times because the `self.continue_workgraph` method can be run concurrently, e.g., in the case of `calcfuntion`, and the `callback` from an Awaitable (submitted process).

## Solution
Add `executed_task` to record the tasks that have already been submitted. We check the `executed_task` before running a task, thus avoiding running tasks multiple times. Since a WorkGraph is run (and only) by one daemon, and Python list `append` operation is atomic, as shown in the [documentation](https://docs.python.org/3/faq/library.html#what-kinds-of-global-value-mutation-are-thread-safe), so the modification on the `executed_tasks` should be thread-safe, within the `asyncio` framework (used by AiiDA Process).

### DIFF
--- a/aiida_workgraph/engine/workgraph.py
+++ b/aiida_workgraph/engine/workgraph.py
@@ -614,9 +614,8 @@ class WorkGraphEngine(Process, metaclass=Protect):
         """Play task."""
         self.report(f"Task {name} action: PLAY.")
 
-    def continue_workgraph(self, exclude: t.Optional[t.List[str]] = None) -> None:
+    def continue_workgraph(self) -> None:
         print("Continue workgraph.")
-        exclude = exclude or []
         self.report("Continue workgraph.")
         # self.update_workgraph_from_base()
         task_to_run = []
@@ -631,7 +630,7 @@ class WorkGraphEngine(Process, metaclass=Protect):
             ]:
                 continue
             ready, output = self.check_parent_state(name)
-            if ready and name not in exclude:
+            if ready and name not in self.ctx.executed_tasks:
                 task_to_run.append(name)
         #
         self.report("tasks ready to run: {}".format(",".join(task_to_run)))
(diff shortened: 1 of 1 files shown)


aiida_workgraph/engine/workgraph.py
```diff
@@ -782,18 +782,28 @@ def run_tasks(self, names: t.List[str], continue_workgraph: bool = True) -> None
                         )
                     )
                     continue
+            # This calcfuntion/workfunction is already run
```
edan-bainglass, line 785: This seems a bit hacky. Why is the work graph in a situation where it is inspecting a calcfunction/workfunction that has already run?

aiida_workgraph/engine/workgraph.py
```diff
@@ -782,18 +782,28 @@ def run_tasks(self, names: t.List[str], continue_workgraph: bool = True) -> None
...
+            if (
+                task["metadata"]["node_type"].upper()
+                in [
+                    "CALCFUNCTION",
+                    "WORKFUNCTION",
+                ]
+                and self.get_task_state_info(name, "state").upper() != "PLANNED"
+            ):
+                continue
```
edan-bainglass: This seems a bit hacky. Why is the work graph in a situation where it is inspecting a calcfunction/workfunction that has already run?

superstar54: Hi Edan, thanks for the comment. Yes, I replaced this by using `executed_tasks`. I also added a detailed description of the bug, and the proposed solution.

edan-bainglass (approved):
LGTM! But best you get another pair of eyes on it, considering the criticality of the bug.

superstar54:
Thanks! I mentioned this in the meeting this morning.  Other people may look at it. I will merge it now to fix the issue for the `calcfuncion`. While for the `atomic` operation and `thread-safe` related, I will keep this in mind.

PR #190, opened 2024-07-19
`PythonJob` supports list of AiiDA node as output

### LINKED ISSUE #189: `PythonJob` supports list of AiiDA node as output.
In the equation of state example, user may return a list of scaled structures, this is not allowed by AiiDA calcfunction.

In the case of `PythonJob`, we may want to support this. A possible solution is to add a new output type: `NamespaceList`.

superstar54:
It would be very helpful (user-friendly) for `PythonJob` to support a list of AiiDA nodes as the output. In order to do that:

- When parsing the list output, give a special name as the key, `list_data_{i}`
- When preparing the submission, if the key of a dict item starts with `list_data_`, we assume that this should convert to a list.

Note that this will be handled internally by the `PythonJob`, so the user will not be aware of this.

Use the example of EOS from @mbercx ,

```python

from aiida_workgraph import task, WorkGraph
from ase import Atoms
from ase.build import bulk

@task.pythonjob(outputs=[{"name": "structure_lst", "identifier": "Namespace"}])
def generate_structures(structure: Atoms, factor_lst: list) -> list[Atoms]:
    structure_lst = []
    for factor in factor_lst:
        structure_strain = structure.copy()
        structure_strain.set_cell(
            structure_strain.cell * factor,
            scale_atoms=True
        )
        structure_lst.append(structure_strain)
    return structure_lst

# user can also decorate a existing function
#generate_structures = task.pythonjob(outputs=[{"name": "structure_lst", "identifier": "Namespace"}])(generate_structures)


atoms = bulk("Si")
wg = WorkGraph("EOS-WG")
generate_structures_task = wg.add_task(
        generate_structures,
        name="generate_structures",
        factor_lst=[0.95, 1.0, 1.05],
        structure=atoms,
    )
wg.submit()

```


superstar54:
This should be handled in the `aiida-core`, see https://github.com/aiidateam/aiida-core/issues/6354 and https://github.com/aiidateam/aiida-core/issues/6799

PR #193, opened 2024-07-31
Adding a prompt to confirm the deletion of a node

### LINKED ISSUE #146: ask confirmation before deleting a workgraph and a data node
The related code for deleting the `workgraph` are

## Backend
[delete_workgraph](https://github.com/aiidateam/aiida-workgraph/blob/a329bf7ce5e30c02afe826038c53508c897e2380/aiida_workgraph/web/backend/app/workgraph.py#L159)

## Frontend

https://github.com/aiidateam/aiida-workgraph/blob/a329bf7ce5e30c02afe826038c53508c897e2380/aiida_workgraph/web/frontend/src/components/WorkGraphTable.js#L156

agoscinski:
Implementing feature request in issue #146


aiida_workgraph/web/frontend/src/components/WorkGraphPrompt.js
```diff
@@ -0,0 +1,55 @@
+// PromptModal.js
+import Modal from 'react-bootstrap/Modal';
+import Button from 'react-bootstrap/Button';
+
+
+function WorkGraphDeleteNodePrompt(props) {
+  return (
+      <Modal 
+        style={{ display: 'block', position: 'initial' }}
```
superstar54: Do not override the default style.

```suggestion

```

aiida_workgraph/web/frontend/src/components/DataNodeTable.js
```diff
@@ -107,8 +110,14 @@ function DataNode() {
...
                             <td>
-                                <button onClick={() => handleDeleteClick(item)} className="action-button delete-button"><FaTrash /></button>
+                                <button onClick={() => setModalShow(true)} className="action-button delete-button"><FaTrash /></button>
                             </td>
+                            <WorkGraphDeleteNodePrompt
+                                show={modalShow}
+                                item={item}
+                                onYesClick={() => handleDeleteConfirmClick(item)}
+                                onNoClick={() => setModalShow(false)}
+                            />
```
superstar54: There are two issues here:

- You set ModalShow to true, this will affect all modal, since you added modal inside each row.
- Better not add a modal inside each row. You can add a new state parameter, `item,` to record the row to be deleted.

agoscinski:
@superstar54 are you okay with these changes? I will implement then workgraph node deletion analogous.

aiida_workgraph/web/frontend/src/components/DataNodeTable.js
```diff
@@ -128,6 +183,13 @@ function DataNode() {
                 breakClassName={'pageBreak'}
             />
             <ToastContainer autoClose={3000} />
+            <WorkGraphConfirmModal
+                show={showConfirmDeleteModal}
+                setShow={setShowConfirmDeleteModal}
+                confirmAction={() => handleDeleteNode(toDeleteItem)}
+                cancelAction={() => {}}
```
superstar54, line 188: This is not needed.
```suggestion

```

aiida_workgraph/web/frontend/src/components/WorkGraphModals.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const handleCancelClick = () => {
+    try {
+      cancelAction();
+    } catch(error) {
+      console.error('While executing the modal cancel action, an action error was raised: ', error)
+    } finally {
+      setShow(false);
+    }
+  };
```
superstar54, line 26: Same here.
```suggestion

```

aiida_workgraph/web/frontend/src/components/DataNodeTable.js
agoscinski: Thanks! now we use a state variable to set the item plus useEffect to detect the change of the state variable that contains the state.

aiida_workgraph/web/frontend/src/components/DataNodeTable.js
```diff
@@ -68,6 +69,51 @@ function DataNode() {
...
+          .then(response => response.json())
+          .then(data => {
+              if (data.deleted_nodes.length > 0) {
+                // the node {item.pk} will be always in the list if the deletion is successfull
+                if (data.deleted_nodes.length > 1) {
+                  let formatted_pks = data.deleted_nodes.map((x) => ` PK<${x.toString()}>`);
+                  data.deleted_nodes.splice(data.deleted_nodes.indexOf(toDeleteItem.pk), 1)
+                  setBodyTextConfirmDeleteModal(
+                    <p>
+                    Are you sure you want to delete node PK&lt;{toDeleteItem.pk}&gt; and {data.deleted_nodes.length} its dependencies?
```
superstar54: Use `dependent nodes` is more clear to the user.
```suggestion
                    Are you sure you want to delete node PK&lt;{toDeleteItem.pk}&gt; and {data.deleted_nodes.length} its dependent nodes.?
```

aiida_workgraph/web/frontend/src/components/DataNodeTable.js
```diff
@@ -68,6 +69,51 @@ function DataNode() {
...
+        if (toDeleteItem != null) {
+          fetch(`http://localhost:8000/api/datanode/delete/${toDeleteItem.pk}?dry_run=True`, {
+              method: 'DELETE',
+          })
+          .then(response => response.json())
+          .then(data => {
+              if (data.deleted_nodes.length > 0) {
+                // the node {item.pk} will be always in the list if the deletion is successfull
+                if (data.deleted_nodes.length > 1) {
+                  let formatted_pks = data.deleted_nodes.map((x) => ` PK<${x.toString()}>`);
```
superstar54: I think only showing the number is enough. Otherwise, there is too much text.
```suggestion
                  let formatted_pks = data.deleted_nodes.map((x) => ` ${x.toString()}`);
```

aiida_workgraph/web/frontend/src/components/DataNodeTable.js
```diff
@@ -68,6 +69,51 @@ function DataNode() {
...
+                  data.deleted_nodes.splice(data.deleted_nodes.indexOf(toDeleteItem.pk), 1)
+                  setBodyTextConfirmDeleteModal(
+                    <p>
+                    Are you sure you want to delete node PK&lt;{toDeleteItem.pk}&gt; and {data.deleted_nodes.length} its dependencies?
+                    <b> A deletion is irreversible.</b>
+                    <br/><br/>
+                    List of items dependent on item PK&lt;{toDeleteItem.pk}&gt; that also will be deleted:
```
superstar54:
```suggestion
                    List of PKs of items dependent on item PK&lt;{toDeleteItem.pk}&gt; that also will be deleted:
```

superstar54 (changes requested):
Thanks for the work! The changes look good to me. I add a few suggestions.

One thing I noted is the style of the table changed. e.g., the color of the header was green before, and now it's white. The style for the table is defined in the `App.css` file, which maybe overrided by the `bootstrap`. Could you change it color back?

aiida_workgraph/web/frontend/src/components/DataNodeTable.js
agoscinski: You are right that for this specific modal in this PR, the `cancelAction` is not needed but I wanted to create a more abstract modal that allows to also clean up stuff in case you allocated resources or changed some state to create the prompt. 

I don't have strong opinions about it, we can always add it back if needed, but it also does not hurt, so what do you think?

aiida_workgraph/web/frontend/src/components/DataNodeTable.js
agoscinski: Maybe we switch to usage of term nodes everywhere?
```suggestion
                    List of PKs of nodes dependent on the node PK&lt;{toDeleteItem.pk}&gt; that also will be deleted:
```

aiida_workgraph/web/frontend/src/components/DataNodeTable.js
agoscinski: I agree that ideally the usage of the `toString` method is better. Reading the message without marking the PKs numbers is not so clear, since you will have numbers that don't correspond to a nodes (like the number of nodes you want to delete). Ideally I would define a class for item that defines the `toString` method. But I would still define in this method to print `PK<item.pk>`

agoscinski: Ah okay it is really only about the list at the end. Yes there I can remove it

aiida_workgraph/web/frontend/src/components/DataNodeTable.js
superstar54: Fine with me if you want to keep it.

agoscinski:
I now gave the workgraph stylesheet distinct names (* ->awg-*) to avoid conflicts with bootstrap.

I also tried to scope bootstrap style sheet using methods described here https://stackoverflow.com/a/46435221 but had no success with them. I think it a bit trickyer for react components.

agoscinski:
I gave the scoping another try, but it seems that the Modal is based on some global style changes that not directly effect the modal elements. At least this is the result I get from scoping
Screenshot 2024-08-13 at 08 08 29

superstar54:
> I now gave the workgraph stylesheet distinct names (* ->awg-*) to avoid conflicts with bootstrap.

Thanks! If the distinct names work, we don't need to use the scoping.

agoscinski:
We managed to fix the issue of the styling of the table by using `th` instead of `td`. Bootstrap was overwriting the `th`style that had priority over the `td`style for the header, so changing the import order was not working before and I went into this rabbit hole of trying out scss.

Now also the deletion prompt for workgraph nodes and frontend tests for both have been implemented.

Will squash merge both commits after successful review.

superstar54 (approved):
LGTM!

PR #199, opened 2024-08-05
Allow nonfunctional usage of decorators

### LINKED ISSUE #191: support decorator without `()`
Current, user need write `()` even though there is not input arguments for the decorator:
```python
@task()
def add(x, y):
    return x+y
```

We should also support the decorator without parentheses, for example:

```python
@task
def add(x, y):
    return x+y
```

```python
@task.calcfunction
def add(x, y):
    return x+y
```

agoscinski:
The decorators that could be used as
```python    
@decorator()
def add(x, y):
    return x+y
```
can now be used nonfunctional
```python
@decorator
def add(x, y):
    return x+y
```
Implements feature request in issue #191


superstar54 (approved):
LGTM!

superstar54:
Hi @agoscinski , two small requests:
- The pre-commit test failed, could you fix it?
- Could you update the doc on this page: https://aiida-workgraph.readthedocs.io/en/latest/concept/task.html#Decorator

superstar54 (approved):

PR #202, opened 2024-08-08
Use uvicorn web server for frontend tests

### LINKED ISSUE #136: Add more test for web application
Blocked by #159

agoscinski:
Since FastAPI's TestClient is not creating any socket, we cannot test the frontend with playwright using a TestClient instance. Therefore for the frontend test I start a uvicorn web server on forked process. We have to start it in a different process as otherwise the loop queue from asyncio conflicts with the one from aiida, so we would not be able to run any jobs. The server is started from pytest so we can inject the pytest aiida profile into it. To have a better control of the profile that the webserver uses I added server settings that can be passed by environment variables.

The PR still needs some work, but the frontend tests work.

EDIT:
Okay it now in good enough shape for a review. I tried to be verbose in the commit messages, but let me know if I should explain things a bit more.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,8 @@ To run the frontend tests in headed mode for you have to set an environment vari
 PYTEST_PLAYWRIGHT_HEADLESS=no pytest -m frontend
 ```
 
+For the frontend tests we start a web server at port `8000`, please free this address for before running the frontend tests.
+
 ### Development on the GUI
 
 For the development on the GUI we use the [REACT](https://react.dev) library
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,63 @@
+
+### Running tests
+
+For running the tests you require only need start rabbitmq in the background.
+The profile that is created during the tests will automatically check its configuration.
+To run all tests use
+
+```console
+pytest
+```
+
+To run only backend tests run
+
+```console
+pytest -m backend
+```
+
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+
+```console
+pytest -m "not backend and not frontend"
+```
+
+To debug the frontend tests you often want to see what happens in the tests.
+By default they are run in headless mode, so no browser is shown.
+To run the frontend tests in headed mode for you have to set an environment variable like this
+```console
+PYTEST_PLAYWRIGHT_HEADLESS=no pytest -m frontend
```
superstar54, line 33: The test seems blocked in the second test (test_workgraph) if I run this headed mode.

superstar54 (approved):
Thanks for the nice work! This is very useful!

The next step would be adding more tests for the backend and frontend.

CONTRIBUTING.md
agoscinski: Ah I see it is because the web server is already started in the CI and this blocks the start of a new one.
https://github.com/aiidateam/aiida-workgraph/blob/76047ee4c77b594193c0891dc4fd68c3012bf1f8/.github/workflows/ci.yaml#L96 
I will remove it to see if the CI works, but in any case I need to add proper error handling from the tests.

agoscinski: Added now a check for the port in tests (see most recent commit).

PR #221, opened 2024-08-13
Add rudimentary frontend tests for all functionalities

### LINKED ISSUE #159: Improve Web APP backend: REST API
The current backend is a minimum working version:
- [x] It loads the profile explicitly, which needs to be avoided.
- [x] Needs more tests.

agoscinski:
For the frontend and backend tests new profiles are created to separate the effects of workgraph tests from the GUI.

Also only one workgraph is run for the frontend tests to save time.

Implements request in issue #159

TODO:
- [x] Move settings tests to mutable section (because the settings stop the daemon at the end, they only change the state if unsuccessful so I keep them where they are)

### DIFF
--- a/tests/web/frontend/conftest.py
+++ b/tests/web/frontend/conftest.py
@@ -1,6 +1,10 @@
 import pytest
+from typing import Generator
 
 from playwright.sync_api import sync_playwright
+from playwright.sync_api import expect
+
+from aiida_workgraph import WorkGraph
 
 import uvicorn
 
@@ -12,6 +16,7 @@ import time
 import os
 import socket
 import errno
+
 
 ################################
 # Utilities for frontend tests #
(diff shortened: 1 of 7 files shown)


superstar54 (changes requested):
Better to move the fixtures to the `conftest.py` file. Could you create two folders for backend and frontend, and then create the conftest.py for each.

agoscinski:
Done. Locally the tests run through.

tests/web/frontend/conftest.py
```diff
@@ -59,6 +65,32 @@ def run_uvicorn_web_server(
 ###############################
 
 
+@pytest.fixture(scope="module")
+def aiida_profile(aiida_config, aiida_profile_factory):
```
agoscinski, line 68: I changed back to aiida_profile as this will overwrite the aiida_profile in the main conftest.py

superstar54 (approved):
Nice work!

PR #223, opened 2024-08-13
Add TaskCollection for waiting_on and children.

### LINKED ISSUE #178: Change the syntax for `wait`
For the moment, set one task wait for others by:

```python
taks3.wait = [task1, taks2]
task3.wait.append(task4)
```

Maybe change to
- wait_for
- add_wait
- clear_up_wait

Besides:
- add validation for the task.
- 
```python
taks3.wait_for(task1)
taks3.wait_for([task1, taks2])
tasks3.waiting_on # return a copy of the `_waiting_on` list, so user can not modify

```
or 
```python
taks3.depends_on([task1, taks2])
```

More detail discussion in this [HackMD note](https://hackmd.io/@superstar54/rJswurIdC)

superstar54:
Although there remains some debate over whether to use the `append` or `replace` method for managing the `waiting_on` list, I've decided to proceed with the `append` approach based on my past experiences. We are open to revising this if there is strong opposition.

For instance, the code below demonstrates how to append "task1" and "task2" to the `waiting_on` list for `task3`, rather than replacing the existing list with ("task1", "task2"):

```python
task3.waiting_on.add(["task1", "task2"])
```

## TaskCollection
Implement a `TaskCollection` to store a collection of tasks. This is used for the `waiting_on` and `children` attributes of the Task class. For example:
```python
self.waiting_on = TaskCollection()
self.children = TaskCollection()
```

## waiting_on
Take `waiting_on` as an example:

```python
# can use the name of the task, or the task object itself
task3.waiting_on.add(["task1", task2])
task3.waiting_on.remove(["task1", "task2"])
task3.waiting_on.clear()
```

When adding or removing the task, it will check if the task exists or not.

## Children
The children attribute is used to assign tasks to a Zone (e.g.,  `While` zone or `If` zone). For example,

```python
# add task1 and task2 to the while loop
while_task.children.add(["task1", "task2"])
```


edan-bainglass:
I agree, regardless of the debate, good to get the code in for live testing. Maybe experience helps (partially) resolve the debate 🤞

edan-bainglass (approved):
LGTM! We can always hash out the API details (naming mostly) down the road, once we have a better feel of the system as a whole.

PR #226, opened 2024-08-14
Add option to set path for python executable for tests

agoscinski:
When running the tests in an environment, the pythonjob tests might fail because the wrong python executable fetched.  The environment gives the possibility to specify the executable for these tests.

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -4,7 +4,7 @@ from typing import Any
 
 
 @pytest.mark.usefixtures("started_daemon_client")
-def test_decorator(fixture_localhost):
+def test_decorator(fixture_localhost, python_executable_path):
     """Test decorator."""
 
     @task.pythonjob(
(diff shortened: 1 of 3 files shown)


tests/test_python.py
```diff
@@ -423,7 +438,13 @@ def make_supercell(atoms: Atoms, dim: int) -> Atoms:
 
     wg = WorkGraph("test_PythonJob_retrieve_files")
     wg.add_task(
-        "PythonJob", function=make_supercell, atoms=atoms, dim=2, name="make_supercell"
+        "PythonJob",
+        function=make_supercell,
+        atoms=atoms,
+        dim=2,
+        name="make_supercell",
+        computer="localhost",
```
agoscinski, line 446: Here I added computer, because it was missing

superstar54 (approved):
LGTM!

PR #229, opened 2024-08-14
Add `PickledFunction` data

### LINKED ISSUE #197: Use a `PickledFunction` class
WorkGraph stores the function using a [dictionary](https://github.com/aiidateam/aiida-workgraph/blob/a329bf7ce5e30c02afe826038c53508c897e2380/aiida_workgraph/utils/__init__.py#L562).

```python
return {
        "executor": pickle.dumps(func),
        "type": "function",
        "is_pickle": True,
        "function_name": func.__name__,
        "function_source_code": function_source_code,
        "function_source_code_without_decorator": function_source_code_without_decorator,
        "import_statements": import_statements,
    }
```

Better to create a `PickledFunction` class for this purpose.

superstar54:
Implement the `PickledFunction` class as a subclass of orm.Data, the class handles 
- storing and retrieving serialized function
- maintaining relevant metadata such as the function name, module name, Python version, and pickle protocol.

PythonJob adds `function` input, which accepts `PickledFunction` as input.

### Docs
Add docs: use the `PythonJob` task outside the WorkGraph to run a Python function on a remote computer. For example, in a `WorkChain` or run a single `CalcJob` calculation.


aiida_workgraph/orm/general_data.py
```diff
@@ -33,53 +37,53 @@ def __str__(self):
...
-        :return: True if using self.get_value returns a reference to the
-            underlying sequence.  False otherwise.
-        :rtype: bool
+        Raises:
+            ImportError: If the pickle protocol does not match.
         """
-        return self.is_stored
+        stored_pickle_protocol = self.base.attributes.get("pickle_protocol")
+
+        if stored_pickle_protocol != cloudpickle.DEFAULT_PROTOCOL:
```
agoscinski: Do you really need to catch this? I remember that python pickling automatically would catch this case, not sure about clouldpickle but I would assume yes

agoscinski (commented):
Have you tried out or thought about using `cloudpickle.register_pickle_by_value` of the imported modules so a installation is not necessary on remote machines?

aiida_workgraph/orm/general_data.py
superstar54: You are right, the clouldpickle will handle this when load the data. I change it to catch the error when unpickling the data, see [2d49a9b](https://github.com/aiidateam/aiida-workgraph/pull/229/commits/2d49a9b82901f3d0c9674fabfd05534090bd1211)

superstar54:
> Have you tried out or thought about using cloudpickle.register_pickle_by_value of the imported modules so a installation is not necessary on remote machines?

Thanks for pointing out this. I wasn't aware of this before. However, I found that this feature is still experimental and may fail in certain cases. We can add this feature as optional. Maybe in another PR.

aiida_workgraph/orm/general_data.py
```diff
@@ -16,12 +19,14 @@ def value(self):
 
 
 class GeneralData(orm.Data):
-    """`Data to represent a pickled value."""
+    """Data to represent a pickled value using cloudpickle."""
+
+    FILENAME = "value.pkl"  # Class attribute to store the filename
```
agoscinski, line 24: So 6his is more like a`PickleObject` data type righ?

agoscinski (approved):
Looks good to me, in the future we might want to be more explicite that we are using cloudpickle, since it works a bit different, but for now I don't see an issue with it. The doc is already transparent about pickling

aiida_workgraph/orm/general_data.py
superstar54: There is no `PickleObject` data type, right? The pickled data is just binary format data, and then we store the binary data in the repository.

superstar54: We can change the name `GeneralData` to `PickledData`, or `AnyData`.

PR #235, opened 2024-08-15
Add `monitor` task decorator to create tasks that to poll some state

### LINKED ISSUE #236: Add `Awaitable` task decorator
To make more precise control over the WorkGraph, we need allow user to add some custom `awaitable` task.
Possible use cases:

Cross-WorkGraph Dependencies: allows adding a task that checks the state of another task inside a different workgraph.
Time trigger: start a task until a given time
Task monitor: monitor the state of a task and determine what's to do for a special event.

superstar54:
This PR introduces an `monitor` task decorator, which can be particularly useful for tasks that need to poll some state (e.g.,  the presence of a file, the state of another WorkGraph) at given intervals until a success criterion is met.

Possible use cases:
- Time trigger: start a task until a given time.
- File trigger: start a task until a file exits.
- Task monitor: monitor the state of a task and determine what's to do for a special event.
- Cross-WorkGraph Dependencies: allows adding a task that checks the state of another task inside a different workgraph.

**Note**: while polling the state, the task will sleep for a given interval (default 1.0 second, can be changed by user), and relinquish control to the WorkGraph engine, so that it can run other tasks.

#### Time monitor
A task wati until a given time.

```python
@task.monitor()
def time_monitor(time):
    """Return True if the current time is greater than the given time."""

# The task will wait until 2024-08-16, 10:54
monitor1 = wg.add_task(time_monitor, time=datetime.datetime(2024, 8, 16, 10, 54, 0))
add1 = wg.add_task(add, "add1",x = 1, y = 2)
add1.waiting_on.add(monitor1)
```

#### File monitor
Start a task until a file exits.

```python
@task.monitor()
def file_monitor(filename):
    """Check if the file exists."""
    import os
    return os.path.exists(filename)

# The task will wait until the file exists, checking every 60 seconds.
monitor1 = wg.add_task(file_monitor, filepath="/tmp/test.txt", interval=60.0)
add1 = wg.add_task(add, "add1",x = 1, y = 2)
add1.waiting_on.add(monitor1)
```

### Builtin tasks
I created some builtin Tasks: `TaskMonitor`,  `TimeMonitor` and `FileMonitor`

#### task monitor
```python
# wait for task "add2" from WorkGraph "test_task_monitor2", you can also use workgraph_pk for a running WorkGraph.
wg1.add_task("workgraph.task_monitor", workgraph_name="test_task_monitor2", task_name="add2")
```
#### time monitor
```python
wg.add_task("workgraph.time_monitor", datetime=datetime.datetime.now() + datetime.timedelta(seconds=10))
```
#### file monitor
```python
wg.add_task("workgraph.file_monitor", filepath="/tmp/test.txt")
```

## General awaitable task
I also created an `awaitable` decorator to allow the user to take full control of the `asyncio` function.

```python
import asyncio

# Load the AiiDA profile
load_profile()

# Define an awaitable task using the new decorator
@task.awaitable()
async def awaitable_func(x, y):
    await asyncio.sleep(0.5)  # Simulate a delay for polling
    return x + y

# Create a WorkGraph and add the awaitable task to it
wg = WorkGraph(name="test_awaitable")
awaitable1 = wg.add_task(awaitable_func, "awaitable_func1", x=1, y=2)

# Run the WorkGraph
wg.run()
```

### About the the asyncio
The awaitable task will let the WorkGraph go to the `Waiting` state. the task will relinquish control to the asyncio event loop, thus the WorkGraph can run other tasks.
However, if there is a long-running calcfunction, the available task will wait for the calcfunction to finish before it can get the control to run the next next step.

## TODO
- How to kill a running monitor task? or trigger the task manually?
- How to keep track of the monitor task, how to set a database record using FunctionNode?


edan-bainglass:
Extensive! Great to have the monitor. I take it this supports monitoring in the AiiDA sense? So currently in AiiDA, a monitor function can be assigned to a `CalcJob` (see [docs](https://aiida.readthedocs.io/projects/aiida-core/en/stable/howto/run_codes.html#how-to-monitor-and-prematurely-stop-a-calculation)). Though initially intended to terminate the job if some condition is met, the monitor function can in principle do more.

Is this PR's feature meant to replicate this via aiida-workgraph methodology? Could you please comment a bit on what it can('t) do? For example, can it kill the job or attach (tracked) outputs to the job? These are currently supported. One interesting future (current?) task for a monitor would be to trigger other jobs if some condition is met. This is a feature request from the W&C community. The trick is how we should keep track of these actions in the provenance.

As for `awaitable` decorator (maybe should be in a separate PR?), no comments there. I don't play with async all that much. However, good to support it, as concurrency is a powerful tool in general!

edan-bainglass (commented):
Extensive! Great to have the monitor. I take it this supports monitoring in the AiiDA sense? So currently in AiiDA, a monitor function can be assigned to a CalcJob (see [docs](https://aiida.readthedocs.io/projects/aiida-core/en/stable/howto/run_codes.html#how-to-monitor-and-prematurely-stop-a-calculation)). Though initially intended to terminate the job if some condition is met, the monitor function can in principle do more.

Is this PR's feature meant to replicate this via aiida-workgraph methodology? Could you please comment a bit on what it can('t) do? For example, can it kill the job or attach (tracked) outputs to the job? These are currently supported. One interesting future (current?) task for a monitor would be to trigger other jobs if some condition is met. This is a feature request from the W&C community. The trick is how we should keep track of these actions in the provenance.

As for awaitable decorator (maybe should be in a separate PR?), no comments there. I don't play with async all that much. However, good to support it, as concurrency is a powerful tool in general!

superstar54:
@edan-bainglass , thanks for for the comment!

The `Calcjob` monitor and the `WorkGraph` monitor are different in terms of the scope of monitoring.

The `Monitor` in AiiDA currently works on the `Calcjob`. It is called within the CalcJob's process, so it can modify its output. This is **intra-process monitoring**.

The `Monitor` in WorkGraph is a task. The monitor task can monitor any events (e.g. time, file changes, etc.) as long as there is a API. Of course, it can also monitor the AiiDA processes using AiiDA API. It can fetch the state of other AiiDA processes, and it can kill/pause/play the process, but it can not modify the data (outputs) of the process. This is **inter-process monitoring**.


> One interesting future (current?) task for a monitor would be to trigger other jobs if some condition is met.

This is exactly what the `monitor` task does.

> The trick is how we should keep track of these actions in the provenance.

The provenance is not tracked as the monitor task is not a aiida process, and it can not be a process, otherwise it will store too much unuseful data. I believe we need to create a special data type for the monitor task to store the provenance, because AiiDA provenance requires AiiDA data,

The `monitor` task is actually implemented using the `awatiable` feature similar to the `awaitable` decorator, but we hide the `asyncio` from the user. The user only needs to write the monitor task as a normal function, and the `WorkGraph` will take care of the rest.

edan-bainglass:
@superstar54 thanks for clarifying the various parts. Let's discuss next week the submission of other jobs via a monitor. This is quite interesting.

As for the PR, good to merge when you're ready.

aiida_workgraph/engine/workgraph.py
```diff
@@ -236,9 +251,15 @@ def _resolve_awaitable(self, awaitable: Awaitable, value: t.Any) -> None:
...
-        )  # remove only if everything went ok, otherwise we may lose track
+        # find awaitable in the self._awaitables with the same pk and remove it
+        for index, a in enumerate(self._awaitables):
+            if a.pk == awaitable.pk:
+                self._awaitables.pop(index)
+                break
+
+        # self._awaitables.remove(
+        # awaitable
+        # )  # remove only if everything went ok, otherwise we may lose track
```
agoscinski: can be removed?

aiida_workgraph/engine/workgraph.py
```diff
@@ -146,7 +148,20 @@ def load_instance_state(
...
         if self._awaitables:
-            # this is a new runner, so we need to re-register the callbacks
+            # this is a new Process instance
+            # For the "ascyncio.tasks.Task" awaitable, because there are only in-memory,
+            # we need to re-run the task
+            should_resume = False
+            for awaitable in self._awaitables:
+                if awaitable.target == "asyncio.tasks.Task":
+                    self._resolve_awaitable(awaitable, None)
+                    self.report(f"reset awatiable task: {awaitable.key}")
```
agoscinski:
```suggestion
                    self.report(f"reset awaitable task: {awaitable.key}")
```

aiida_workgraph/engine/workgraph.py
```diff
@@ -146,7 +148,20 @@ def load_instance_state(
         self.set_logger(self.node.logger)
 
         if self._awaitables:
-            # this is a new runner, so we need to re-register the callbacks
+            # this is a new Process instance
+            # For the "ascyncio.tasks.Task" awaitable, because there are only in-memory,
+            # we need to re-run the task
```
agoscinski:
```suggestion
            # For the "ascyncio.tasks.Task" awaitable, because there are only in-memory,
            # we need to re-run the task and by that creating a new process instance
```
Is this what you mean? Maybe you can connect the two sentences in some logical way.

aiida_workgraph/engine/workgraph.py
```diff
@@ -146,7 +148,20 @@ def load_instance_state(
...
+            should_resume = False
+            for awaitable in self._awaitables:
+                if awaitable.target == "asyncio.tasks.Task":
+                    self._resolve_awaitable(awaitable, None)
```
agoscinski, line 156: when reading `_resolve_awaitable` I saw that the docstring for the parameter`value` is missing https://github.com/aiidateam/aiida-workgraph/blob/b54fd5c8145ac37975b33d980d131b7cb266e5c5/aiida_workgraph/engine/workgraph.py#L229-L231

aiida_workgraph/engine/workgraph.py
```diff
@@ -146,7 +148,20 @@ def load_instance_state(
...
+                    self.report(f"reset awatiable task: {awaitable.key}")
+                    self.reset_task(awaitable.key)
+                    should_resume = True
+            if should_resume:
+                self._update_process_status()
+                self.resume()
+            # For other awatiable, because they exist in the db, we only need to re-register the callbacks
```
agoscinski:
```suggestion
            # For other awaitables, because they exist in the db, we only need to re-register the callbacks
```

aiida_workgraph/engine/workgraph.py
```diff
@@ -369,22 +394,34 @@ def _on_awaitable_finished(self, awaitable: Awaitable) -> None:
         """
         print("on awaitable finished: ", awaitable.key)
         self.logger.info(
-            "received callback that awaitable %d has terminated", awaitable.pk
+            "received callback that awaitable %d has terminated", awaitable.key
```
agoscinski: pk is the AiiDA Process identifier, and key is the Awaitable identifier? Do you think it might be useful to put in both information? Something like
```suggestion
            "received callback that awaitable %d started in process pk<%d> has terminated", awaitable.key, awaitable.pk
```

tests/test_awaitable_task.py
```diff
@@ -0,0 +1,89 @@
...
+            await asyncio.sleep(0.5)
+        return x + y
+
+    wg = WorkGraph(name="test_awaitable")
+    awaitable1 = wg.add_task(awaitable_func, "awaitable_func1", x=1, y=2)
+    add1 = wg.add_task(decorated_add, "add1", x=1, y=awaitable1.outputs["result"])
+    wg.run()
+    report = get_workchain_report(wg.process, "REPORT")
+    assert "Waiting for child processes: awaitable_func1" in report
+    assert add1.outputs["result"].value == 4
```
agoscinski: more expressive names are always nicer to read
```suggestion
    awaitable_func_task = wg.add_task(awaitable_func, "awaitable_func_task", x=1, y=2)
    add_task = wg.add_task(decorated_add, "add_task", x=1, y=awaitable_task.outputs["result"])
    wg.run()
    report = get_workchain_report(wg.process, "REPORT")
    assert "Waiting for child processes: awaitable_func_task" in report
    assert add_task.outputs["result"].value == 4
```

tests/test_awaitable_task.py
```diff
@@ -0,0 +1,89 @@
...
+        "workgraph.time_monitor",
+        "monitor1",
+        datetime=datetime.datetime.now() + datetime.timedelta(seconds=10),
+    )
+    add1 = wg.add_task(decorated_add, "add1", x=1, y=2)
+    add1.waiting_on.add(monitor1)
+    wg.run()
+    report = get_workchain_report(wg.process, "REPORT")
+    assert "Waiting for child processes: monitor1" in report
+    assert add1.outputs["result"].value == 3
```
agoscinski, line 39:
```suggestion
    monitor_task = wg.add_task(
        "workgraph.time_monitor",
        "monitor_task",
        datetime=datetime.datetime.now() + datetime.timedelta(seconds=10),
    )
    add_task = wg.add_task(decorated_add, "add_task", x=1, y=2)
    add_task.waiting_on.add(monitor1)
    wg.run()
    report = get_workchain_report(wg.process, "REPORT")
    assert "Waiting for child processes: monitor_task" in report
    assert add_task.outputs["result"].value == 3
```

tests/test_awaitable_task.py
```diff
@@ -0,0 +1,89 @@
...
+    monitor1 = wg.add_task(
+        "workgraph.file_monitor", name="monitor1", filepath="/tmp/test_file_monitor.txt"
+    )
+    wg.add_task(create_test_file, "create_test_file1")
+    add1 = wg.add_task(decorated_add, "add1", x=1, y=2)
+    add1.waiting_on.add(monitor1)
+    wg.run()
+    report = get_workchain_report(wg.process, "REPORT")
+    assert "Waiting for child processes: monitor1" in report
+    assert add1.outputs["result"].value == 3
```
agoscinski: pytest has the `tmp_path` fixture that is always autouse that cleans up the file https://docs.pytest.org/en/stable/how-to/tmp_path.html after the run of the test.
```suggestion
def test_file_monitor(decorated_add, tmp_path):
    """Test the file monitor task."""

    monitor_file_path = tmp_path / "test_file_monitor.txt"
    @task.awaitable()
    async def create_test_file(filepath=monitor_file_path, t=2):
        await asyncio.sleep(t)
        with open(filepath, "w") as f:
            f.write("test")

    wg = WorkGraph(name="test_file_monitor")
    monitor_task = wg.add_task(
        "workgraph.file_monitor", name="monitor_task", filepath=monitor_file_path
    )
    wg.add_task(create_test_file, "create_test_file_task")
    add_task = wg.add_task(decorated_add, "add_task", x=1, y=2)
    add_task.waiting_on.add(monitor_task)
    wg.run()
    report = get_workchain_report(wg.process, "REPORT")
    assert "Waiting for child processes: monitor_task" in report
    assert add_task.outputs["result"].value == 3
```

tests/test_awaitable_task.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    add1 = wg2.add_task(decorated_add, "add1", x=1, y=2, t=0)
+    add1.waiting_on.add(monitor1)
+    wg2.submit()
+    #
+    wg1 = WorkGraph(name="test_task_monitor1")
+    add1 = wg1.add_task(decorated_add, "add1", x=1, y=2, t=5)
+    wg1.submit(wait=True)
+    wg2.wait()
+    assert wg2.tasks["add1"].node.ctime > wg1.tasks["add1"].node.ctime
```
agoscinski: Naming again and a bit doc to differ it from above test. 
```suggestion
def test_task_monitor(decorated_add):
    """Test the file monitor task referring to tasks of other workgraphs."""
    wg2 = WorkGraph(name="wg2")
    monitor_task = wg2.add_task(
        "workgraph.task_monitor",
        name="monitor_task",
        workgraph_name="wg1",
        task_name="add_task",
    )
    wg2_add_task = wg2.add_task(decorated_add, "add_task", x=1, y=2, t=0)
      add_task.waiting_on.add(monitor1)
    wg2.submit()
    
    # wg2 waits now till wg1 finishes its "add_task", so we create it
    wg1 = WorkGraph(name="wg1")
    wg1_add_task = wg1.add_task(decorated_add, "add_task", x=1, y=2, t=5)
    wg1.submit(wait=True)
    wg2.wait()
    assert wg2.tasks["add_task"].node.ctime > wg1.tasks["add_task"].node.ctime
```

aiida_workgraph/engine/workgraph.py
```diff
@@ -646,9 +682,27 @@ def update_task_state(self, name: str) -> None:
             self.task_set_context(name)
             self.report(f"Task: {name} finished.")
         else:
-            task["results"] = None
+            task.setdefault("results", None)
+
+        self.update_parent_task_state(name)
 
+    def set_normal_task_results(self, name, results):
```
agoscinski, line 693: Is normal task somewhere explained in the docstring? If not could you explain it somewhere? It becomes clear from reading the code, but this should be also somewhere documented

aiida_workgraph/engine/workgraph.py
```diff
@@ -646,9 +682,27 @@ def update_task_state(self, name: str) -> None:
...
+                return self.exit_codes.OUTPUS_NOT_MATCH_RESULTS
+            for i in range(len(task["outputs"])):
+                task["results"][task["outputs"][i]["name"]] = results[i]
+        elif isinstance(results, dict):
+            task["results"] = results
+        else:
+            task["results"][task["outputs"][0]["name"]] = results
+        self.set_task_state_info(name, "state", "FINISHED")
+        self.task_set_context(name)
+        self.report(f"Task: {name} finished.")
```
agoscinski: I know a bit nitpicky, but shouldn't the order be?
```suggestion
        self.task_set_context(name)
        self.set_task_state_info(name, "state", "FINISHED")
        self.report(f"Task: {name} finished.")
```

I don't know if the task should be marked as finished if task_set_context fails, but this is the pattern I see here

aiida_workgraph/engine/workgraph.py
```diff
@@ -646,9 +682,27 @@ def update_task_state(self, name: str) -> None:
...
         self.update_parent_task_state(name)
+        print("set normal task results: ", name, "results: ", task["results"])
```
agoscinski: Should this be not in the logged?
```suggestion
        self.logger.info("set normal task results: ", name, "results: ", task["results"])
```

aiida_workgraph/engine/workgraph.py
```diff
@@ -1066,40 +1119,74 @@ def run_tasks(self, names: t.List[str], continue_workgraph: bool = True) -> None
...
                 self.continue_workgraph()
+            elif task["metadata"]["node_type"].upper() in ["AWAITABLE"]:
+                for key in self.ctx._tasks[name]["metadata"]["args"]:
+                    kwargs.pop(key, None)
+                awaitable_target = asyncio.ensure_future(
+                    self.run_executor(executor, args, kwargs, var_args, var_kwargs),
+                    loop=self.loop,
+                )
+                awaitable = self.construct_awaitable_function(name, awaitable_target)
+                self.set_task_state_info(name, "state", "RUNNING")
```
agoscinski, line 1150: If we report the finished state should we not also log somewhere the running state? 
```suggestion
                self.set_task_state_info(name, "state", "RUNNING")
                self.report(f"Task: {name} is running.", )
```
Maybe on a different log level like debug, but for that I think one needs to write a new function, but the report function is fairly easy to implement. https://github.com/aiidateam/aiida-core/blob/fb3686271fcdeb5506838a5a3069955546b05460/src/aiida/engine/processes/process.py#L589-L600

aiida_workgraph/engine/workgraph.py
```diff
@@ -1066,40 +1119,74 @@ def run_tasks(self, names: t.List[str], continue_workgraph: bool = True) -> None
...
+                for key in self.ctx._tasks[name]["metadata"]["args"]:
+                    kwargs.pop(key, None)
+                # add function and interval to the args
+                args = [executor, kwargs.pop("interval", 1), *args]
+                awaitable_target = asyncio.ensure_future(
+                    self.run_executor(monitor, args, kwargs, var_args, var_kwargs),
+                    loop=self.loop,
+                )
+                awaitable = self.construct_awaitable_function(name, awaitable_target)
+                self.set_task_state_info(name, "state", "RUNNING")
```
agoscinski, line 1163: same question as above

agoscinski (changes requested):
Thanks for the PR, looks good to me. Only minor suggestions.

tests/test_awaitable_task.py
superstar54: Thanks! I renamed `awaitable1` to `awaitable_func1`. Generally, I recommend not using `_task` in the task's name, because
- sometimes, the task name is already very long, so we just skip `_task.`
- In the GUI, it will show the name of the task, we try to avoid showing `_task` on the GUI. because we know it's a task.
- A `suffix` is recommend to add to the name, because, we many create multiple tasks from the same Task class.

tests/test_awaitable_task.py
superstar54: See the previous comment regarding the task's name.

aiida_workgraph/engine/workgraph.py
superstar54: At the top of the `run_tasks` function, it reports that to run the task and also shows the type of the task. Thus I prefer not to report the `running` state.

aiida_workgraph/engine/workgraph.py
superstar54: Thanks! I added one line to the docstring of this function.
```
A  normal task is created by decorating a function with @task()
```

tests/test_awaitable_task.py
superstar54: I didn't know this. Thanks!

aiida_workgraph/engine/workgraph.py
superstar54: Thanks! Done!

aiida_workgraph/engine/workgraph.py
superstar54: I think it doesn't need to print out it anymore, so I deleted it.

aiida_workgraph/engine/workgraph.py
superstar54: Yes, done!

aiida_workgraph/engine/workgraph.py
superstar54: thanks! I revised the text.

aiida_workgraph/engine/workgraph.py
superstar54: See previous reply.

aiida_workgraph/engine/workgraph.py
superstar54: I put both there if the pk and key are not the same.

agoscinski (approved):
Thanks looks good!

PR #238, opened 2024-08-18
Show timeline of called_process

superstar54:
In the while zone, tasks repeat several times, however, in the timeline it only shows the time for the last iteration of the task. As show below, there is some missing info between `add1` and `add2`.

Screenshot from 2024-08-18 14-12-06

This PR adds the `called_process` option. Switching to this option, it will show the timeline for all the processes that called by the WorkGraph:
Screenshot from 2024-08-18 14-12-18

### DIFF
--- a/aiida_workgraph/utils/__init__.py
+++ b/aiida_workgraph/utils/__init__.py
@@ -261,7 +261,7 @@ def get_parent_workgraphs(pk: int) -> list:
 
 
 def get_processes_latest(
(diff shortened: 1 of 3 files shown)


superstar54:
Hi @agoscinski , the frontend test caught a bug in this PR. https://github.com/aiidateam/aiida-workgraph/actions/runs/10441999297/job/28914107364

Very helpful!

PR #243, opened 2024-08-19
Migrate quick_start to sphinx-gallery

agoscinski:
Migrate the `quck_start.ipynb` to sphinx-gallery:

- create a `base` conda env, and install aiida core services, including the rabbitmq.
- create a profile using `verdi presto`
- convert the graphviz to an SVG XML string, then embed it into an HTML file

After setting up the AiiDA environment, migration of other notebooks becomes straightforward.


.readthedocs.yaml
```diff
@@ -3,10 +3,24 @@ version: 2
...
   commands:
-    - python -mvirtualenv $READTHEDOCS_VIRTUALENV_PATH
+    - which python
+    - conda init
+    - echo $SHELL
+    - python --version
+    - which conda 
+    - conda --version
+    - cat /home/docs/.bashrc
+    - . /home/docs/.bashrc
```
superstar54: Is `.bashrc` executable? maybe try `source /home/docs/.bashrc`

PR #249, opened 2024-08-21
Test/rtd rabbitmq

superstar54:


superstar54:
Confirmed that we need to sleep a few seconds after starting the rabbitmq server.

PR #253, opened 2024-08-22
create and submit the process using `self.submit`

### LINKED ISSUE #252: WorkGraph generated by Graph builder stuck in `Created` status
See #251 

The engine submit the workgraph generated by graph_builder task in two steps:
- run `wg.save()` to create a process
- to submit the process using `self.submit`

the problem is that the `runner` may not be the same in the two steps, thus the process is lost.

superstar54:
Fix #252 

Instead of creating and submitting the process in two steps, this PR use the `self.submit` only.


agoscinski (approved):
Works! It works in a ipynb also without adding additional commands!

Issue #195, opened 2024-07-31
Section graph_builder mentions "dynamic" creation but other sections introduce them


agoscinski:
In the doc https://aiida-workgraph.readthedocs.io/en/latest/howto/graph_builder.html? The term "dyanmic" is used, but the "dynamic" graphs are shown in different sections https://aiida-workgraph.readthedocs.io/en/latest/howto/if.html One could just change the doctree structure
```
graph_builder
 |------ if
 |_____ while
```

superstar54:
A control flow doc is added in the howto, and the `Graph` task is moved to `concepts. Hence I close this issue.

PR #277, opened 2024-08-27
Create subsection "Nested workflow" in the docs 

agoscinski:
Moves the graph builder, if task and while task examples into this subsection. Also removed the graph_builder.ipynb since it is now generated. Solves issue #195


agoscinski:
`test_pause_task_after_submit` seems a bit unstable it failed already two times in different jobs

superstar54 (changes requested):
Looking at the `if` and `while` task again, because we now have new `If` task and `while` task, thus there are two ways to implement `if`
- `if` task, **dose not need nested workgraph**
- use `graph_builder`, thus nested workgraph.

thus it's better to keep `if` and `while` separately, and not move to the `nested_workgraph`.

I suggest at the end of the `graph_builder.py`, we make a link to show the application in the `if` and `while`.

agoscinski:
I dont think the graph_builder example is really about the dynamic part as it just forwards inputs like any task does it. We could make another example that runs different tasks depending on the inputs. Something like
```python
@task.calcfunction()
def add_one(x):
    return x.value+1

@task.calcfunction()
def modulo_five(x):
    return x % 5

@graph_builder(outputs = [...])
def my_modular(i: orm.Int):
    wg = WorkGraph()
    if i.value < 5:
        task = wg.add_task(add_one)
    else:
        task = wg.add_task(modulo_five)
    # need wg.selector to expose for linking?
    return wg
```
Then one could say that this does not preserve provenance and directly interlude to the `If` workgraph

agoscinski:
Added an example for a dynamic use case of the graph_builder

docs/gallery/howto/dynamic_workflows/autogen/dynamic_graph_builder.py
```diff
@@ -0,0 +1,70 @@
...
+def add_modulo(i: Int):
+    wg = WorkGraph()
+    if i.value < 2:
+        task = wg.add_task(add_one, x=i)
+    else:
+        task = wg.add_task(modulo_two, x=i)
+
+    task.set_context({"result": "out"})
+    return wg
+
```
superstar54: This is a duplicate as in the `if.` I would suggest using a for loop inside the graph_builder so that the number of tasks depends on an input value, making the work graph dynamic.

superstar54 (changes requested):
Hi @agoscinski , Thanks for the work.

I would suggest not creating a `dynamic-workflows` section, but moving all three notebooks into the `howto`, so that user can see the `dynamic`, `if` and `while` immediately.

superstar54:
> Then one could say that this does not preserve provenance and directly interlude to the If workgraph

Why ti does not preserve provenance?

docs/gallery/howto/dynamic_workflows/autogen/dynamic_graph_builder.py
agoscinski: I think we have quite a few for loops in the examples too therefore I don't see duplication as a reason to rather use for loops. I feel this example is more educative because you are changing the type of task, you are in different branch of your program depending on your input which might be more intuitive to be dynamic.

But I can also just add both.

agoscinski:
> I would suggest not creating a dynamic-workflows section, but moving all three notebooks into the howto, so that user can see the dynamic, if and while immediately.

Okay, but the howto's starting to get messy and need some structure at some point, but I think for now one still can put everything there

agoscinski:
> Why ti does not preserve provenance?

I mean more that it does not store provenance as transparently as using the `If` WorkGraph, but looking at the provenance graph, It is not really capturing the if-then-else logic, but just only some additional information about the condition. But at least in the GUI the if-then-else is capture more transparently.

agoscinski:
I merged now the nested and dynamic example because it seemed strange to see them separate

docs/source/howto/index.rst
```diff
@@ -9,9 +9,10 @@ This section contains a collection of HowTos for various topics.
    :caption: Contents:
 
    autogen/graph_builder
-   autogen/parallel
    if
    while
+   dynamic_workflows/index
```
superstar54: there is no dynamic_workflows/index file
```suggestion

```

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -1,47 +1,28 @@
...
 # %%
-# Introduction
-# ============
-# The `Graph Builder` allow user to create a dynamic workflow based on the input value, as well as nested workflows.
-#
-# Load the AiiDA profile.
-
+# Nested workflows
+# ================
+# The `Graph Builder` allow user to create nested workflows from an input.
```
superstar54: Better to give an overview: user can create nested WorkGraph in two ways:
- Create a Task from the workgraph 
- Use graph builder.

Also discuss what's the upside and downside of the two approaches, or later in each section.

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -55,28 +36,78 @@ def multiply(x, y):
...
+    wg.add_task(
+        add_multiply(x=add_multiply1.outputs["multiply.result"], y=Int(3), z=Int(4))
+    )
+except Exception as err:
+    print(err)
+
+# %%
+# For that use case we need to use the graph builder
+
+# Create a graph builder function
```
superstar54: Remove this empty line, otherwise the format will be wrong.

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -1,47 +1,28 @@
...
-#
-#     @task.graph_builder(outputs = [{"name": "multiply", "from": "multiply.result"}])
-#
-# This will expose the `result` output of the `multiply` task as the `multiply` output of the WorkGraph.
-#
-#
-
+# -------
+# Suppose we want a WorkGraph which includes another WorkGraph`(x+y)*z` inside it.
+# We can actually add a WorkGraph to another WorkGraph
```
superstar54: This belongs to the `Create a Task from the workgraph`

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -55,28 +36,78 @@ def multiply(x, y):
...
+
+wg = WorkGraph("nested_workgraph")
+add_multiply1 = wg.add_task(add_multiply(x=Int(2), y=Int(3), z=Int(4)))
+
+try:
+    wg.add_task(
+        add_multiply(x=add_multiply1.outputs["multiply.result"], y=Int(3), z=Int(4))
+    )
+except Exception as err:
+    print(err)
```
superstar54: Actually, one can link the two WorkGraph tasks. But we need to write the code in a different way:
```python
def add_multiply(x=None, y=None, z=None):
    wg = WorkGraph()
    wg.add_task(add, name="add", x=x, y=y)
    wg.add_task(multiply, name="multiply", x=z)
    wg.add_link(wg.tasks["add"].outputs[0], wg.tasks["multiply"].inputs["y"])
    return wg

wg = WorkGraph("nested_workgraph")
add_multiply1 = wg.add_task(add_multiply(x=2, y=3, z=4), name="add_multiply1")
add_multiply2 = wg.add_task(add_multiply(y=5, z=6), name="add_multiply2")
wg.add_link(add_multiply1.outputs["multiply.result"], add_multiply2.inputs["add.x"])
wg.run()
```

The difference between the above code and a `graph_builder` task is that in `graph_builder`, the workgraph is dynamic and only created during execution, while the above workgraph is static, so that we can access the socket directly, e.g., `add.x`.
Both can used to create nested workgrpah, but `graph_builder` is a black box, which is the downside. The upside is that it allows dynamic workgraph generation.

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -108,15 +146,15 @@ def add_multiply(x, y, z):
 
 
 # %%
-# Create the a Task from the workgraph (Experimental)
-# ---------------------------------------------------
+# Create a Task from the workgraph (Experimental)
+# -----------------------------------------------
```
superstar54: Since you mentioned this part at the beginning, so better to move this section before `graph_builder`.

superstar54 (changes requested):
Hi @agoscinski , thanks for the update!

The `dynamic_graph_builder.py` is not used, I think you want to remove it.

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -55,30 +48,84 @@ def multiply(x, y):
...
+    wg.add_link(wg.tasks["add"].outputs[0], wg.tasks["multiply"].inputs["y"])
+    return wg
+
+
+wg = WorkGraph("nested_workgraph")
+# Creating a task from the WorkGraph
+add_multiply1 = wg.add_task(add_multiply(x=Int(2), y=Int(3), z=Int(4)))
+add_multiply2 = wg.add_task(add_multiply(x=Int(2), y=Int(3)))
+# link the output of a task to the input of another task
+wg.add_link(add_multiply1.outputs[0], add_multiply2.inputs["add.z"])
```
agoscinski:
```suggestion
wg.add_link(add_multiply1.outputs[0], add_multiply2.inputs["multiply.z"])
```

docs/gallery/howto/autogen/graph_builder.py
agoscinski: Have done something similar

docs/gallery/howto/autogen/graph_builder.py
agoscinski: It is at the beginning now

docs/gallery/howto/autogen/graph_builder.py
agoscinski: Okay used that example and explained a bit

docs/gallery/howto/autogen/graph_builder.py
agoscinski: Is now integrated into the first part

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -55,30 +48,84 @@ def multiply(x, y):
...
+    wg.add_link(wg.tasks["add"].outputs[0], wg.tasks["multiply"].inputs["y"])
+    return wg
+
+
+wg = WorkGraph("nested_workgraph")
+# Creating a task from the WorkGraph
+add_multiply1 = wg.add_task(add_multiply(x=Int(2), y=Int(3), z=Int(4)))
+add_multiply2 = wg.add_task(add_multiply(x=Int(2), y=Int(3)))
+# link the output of a task to the input of another task
+wg.add_link(add_multiply1.outputs[0], add_multiply2.inputs["multiply.z"])
```
agoscinski:
```suggestion
wg.add_link(add_multiply1.outputs[0], add_multiply2.inputs["multiply.y"])
```

docs/gallery/howto/autogen/graph_builder.py
agoscinski: Hm.. it is not properly working
```python
# define add task
@task.calcfunction()
def add(x, y):
    return x + y


# define multiply task
@task.calcfunction()
def multiply(x, y):
    return x * y


def add_multiply(x=None, y=None, z=None):
    wg = WorkGraph()
    wg.add_task(add, name="add", x=x, y=y)
    wg.add_task(multiply, name="multiply", x=z)
    wg.add_link(wg.tasks["add"].outputs[0], wg.tasks["multiply"].inputs["y"])
    return wg


wg = WorkGraph("nested_workgraph")
# Creating a task from the WorkGraph
add_multiply1 = wg.add_task(add_multiply(x=Int(2), y=Int(3), z=Int(4)))
add_multiply2 = wg.add_task(add_multiply(x=Int(2), y=Int(3)))
# link the output of a task to the input of another task
wg.add_link(add_multiply1.outputs[0], add_multiply2.inputs["multiply.x"])
wg.to_html()

# %%
# Run the workgraph

wg.run()
```
Gives
```
Error: Error in task multiply: Cannot convert value of type <class 'aiida.orm.utils.managers.NodeLinksManager'> to AiiDA type.
```

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -55,30 +48,84 @@ def multiply(x, y):
...
+    wg.add_link(wg.tasks["add"].outputs[0], wg.tasks["multiply"].inputs["y"])
+    return wg
+
+
+wg = WorkGraph("nested_workgraph")
+# Creating a task from the WorkGraph
+add_multiply1 = wg.add_task(add_multiply(x=Int(2), y=Int(3), z=Int(4)))
+add_multiply2 = wg.add_task(add_multiply(x=Int(2), y=Int(3)))
+# link the output of a task to the input of another task
+wg.add_link(add_multiply1.outputs[0], add_multiply2.inputs["multiply.y"])
```
superstar54:
```suggestion
wg.add_link(add_multiply1.outputs["multiply.result"], add_multiply2.inputs["multiply.x"])
```

agoscinski: Thanks!

docs/gallery/howto/autogen/graph_builder.py
agoscinski: Okay now it works!

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -91,79 +142,125 @@ def add_multiply(x, y, z):
...
-# link the output of a task to the input of another task
-wg2.add_link(add_multiply1.outputs["multiply.result"], add_multiply2.inputs["add.x"])
+@task.graph_builder(outputs=[{"name": "result", "from": "context.task_out"}])
+def for_loop(nb_iterations: Int):
+    wg = WorkGraph()
+    for i in range(nb_iterations.value):
+        task = wg.add_task(add_one, x=i)
+
+    # We cannot refer to a specific task as output in the graph builder decorator
+    # as in the examples below as the name of the last task depends on the input.
```
agoscinski:
```suggestion
    # as in the examples before since the name of the last task depends on the input.
    # Remember that each task is always assigned unique name automatically.
```

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -91,79 +142,125 @@ def add_multiply(x, y, z):
 # %%
 # Generate node graph from the AiiDA process,and we can see that the `multiply` task is executed.
```
agoscinski:
```suggestion
# Generate node graph from the AiiDA process,and we can see that the `multiply` task was executed.
```

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -91,79 +142,125 @@ def add_multiply(x, y, z):
...
 
-
-from aiida_workgraph.utils import generate_node_graph
-
 generate_node_graph(wg.pk)
 
+# %%
+# Looking at the process list we can also that multiple WorkGraphs have been submitted
```
agoscinski:
```suggestion
# Looking at the process list we can also that multiple WorkGraphs have been submitted.
# Please run this now in the terminal:
```

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -91,79 +142,125 @@ def add_multiply(x, y, z):
...
-
 wg = add_multiply(2, 3, 4)
 wg.submit(wait=True)
 
 
 # %%
-# Create the a Task from the workgraph (Experimental)
-# ---------------------------------------------------
-# One can create a Task from a WorkGraph directly.
+# More usage (like `if` and `while`) of graph builder will be shown in the following tutorials.
```
agoscinski:
```suggestion
# More usage (like `if` and `while`) of graph builder will be shown in the following how-tos.
```

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -91,79 +142,125 @@ def add_multiply(x, y, z):
...
 
 
-# create a task using the WorkGraph
-print("Inputs:")
-for input in add_multiply1.inputs:
-    print(f"  - {input.name}")
-print("Outputs:")
-for output in add_multiply1.outputs:
-    print(f"  - {output.name}")
+wg = WorkGraph()
```
superstar54: We need to add a name to the WorkGraph; otherwise, the HTML file will be overridden.
```suggestion 
wg = WorkGraph("Nested workflow: For")
```

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -91,79 +142,125 @@ def add_multiply(x, y, z):
...
+        task = wg.add_task(modulo_two, x=i)
 
-from aiida_workgraph.utils import generate_node_graph
+    # same concept as before, please read the for loop example for explanation
+    task.set_context({"result": "task_out"})
+    return wg
 
-generate_node_graph(wg2.pk)
 
+wg = WorkGraph()
```
superstar54:
```suggestion
wg = WorkGraph("Nested workflow: If)
```

superstar54 (changes requested):

superstar54 (approved):
LGTM, thanks!

PR #283, opened 2024-08-28
Use calcfunction for getitem in `For` loop workgraph

agoscinski:
To track the provenance in the for loop we now use a calcfunction to track provenance for the `For` loop workgraph. I would like to allow more types than integers for the sequence, but I am not sure how to generically test for types that are supported by aiida core. Some orm types have a `_type` class attribute, but not all of them.

TODO:
- [ ] add notebook reason why to choose this over a normal for loop (less tasks created), or at least move this to an issue

### DIFF
--- a/aiida_workgraph/engine/workgraph.py
+++ b/aiida_workgraph/engine/workgraph.py
@@ -22,6 +22,7 @@ from aiida.orm.utils.serialize import deserialize_unsafe, serialize
 
 from aiida.engine.processes.exit_code import ExitCode
 from aiida.engine.processes.process import Process
+from aiida.engine import calcfunction
 
 from aiida.engine.processes.workchains.awaitable import (
     Awaitable,
@@ -30,7 +31,7 @@ from aiida.engine.processes.workchains.awaitable import (
     construct_awaitable,
 )
 from aiida.engine.processes.workchains.workchain import Protect, WorkChainSpec
-from aiida.engine import run_get_node
+from aiida.engine import calcfunction, run_get_node
 from aiida_workgraph.utils import create_and_pause_process
 from aiida_workgraph.task import Task
 from aiida_workgraph.utils import get_nested_dict, update_nested_dict
@@ -920,7 +921,18 @@ class WorkGraphEngine(Process, metaclass=Protect):
         if should_run:
             self.reset()
             self.set_tasks_state(condition_tasks, "SKIPPED")
-            self.ctx["i"] = self.ctx._sequence[self.ctx._count]
+
+            @calcfunction
+            def getitem(sequence, count):
+                value = sequence.get_list()[count.value]
+                # TODO can we make this more generic?
+                if isinstance(value, int):
+                    return orm.Int(value)
+                else:
+                    raise TypeError("`sequence` only supports sequence of integers")
+
+
+            self.ctx["i"] = getitem(self.ctx._sequence, self.ctx._count)
         self.ctx._count += 1
         return should_run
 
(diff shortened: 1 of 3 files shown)


superstar54:
Hi @agoscinski , thanks for the work.

There is a function `orm.to_aiida_type` which can serialize the data automatically, and you can check which Python data types it support.


It's also would be good if the user can supply a list of AiiDA data as the input for `sequence`, e.g.,
```python
[structure0, structure1, ...]
```

Because AiiDA does not support a list of AiiDA data as input, internally, we can change it to a dict with keys, the keys keep the order information.
```python
{"sequence_0": structure0, ...}
```

and then, in the WorkGraph engine, if the sequence is a dict, we get the AiiDA data directly without serialization.

agoscinski:
Thanks @superstar54 applied your suggestion. The test pass, but it does not test the other use cases you mentioned. I tried to implement a dictionary, but it is a bit opaque to me how to change the iterator to be compatible with dictionaries that use hashable keys. It seems the iterator is hard coded to only work for collections that can be accessed by incremental increasing numbers like lists.  Maybe we should generalize this for arbitrary generators. Not sure if this PR is the right place.

aiida_workgraph/engine/workgraph.py
```diff
@@ -920,7 +920,19 @@ def check_for_conditions(self) -> bool:
         if should_run:
             self.reset()
             self.set_tasks_state(condition_tasks, "SKIPPED")
-            self.ctx["i"] = self.ctx._sequence[self.ctx._count]
+
+            @calcfunction
+            def __getitem__(sequence, count):
+                value = sequence[count.value]
+                # only convert value f not already orm type because sequence
```
superstar54:
```suggestion
                # only convert value if not already orm type because sequence
```

superstar54 (changes requested):
Hi @agoscinski , thanks for the work! 

Please test for the changes.

As mentioned in my previous comment, if you use a list of orm.data as the sequence, it will raise an error, because we need to change the list to a dict.

Therefore, you need also modify the `aiida_workgraph.workgraph.WorkGraph` to convert it into a dict. Then inside the engine, the sequence is a dict.

agoscinski:
> As mentioned in my previous comment, if you use a list of orm.data as the sequence, it will raise an error, because we need to change the list to a dict.

That is for non json serializable data types like `orm.StructureData` for others it will automatically convert it and is not a problem. My problem with the dynamic workaround is that in this case it is shown differently in the provenance graph and also will create a lot of links.

That was the initial idea
```python
@calcfunction
def __getitem__(seq, idx):
    return orm.to_aiida_type(seq[idx.value])

node = __getitem__([1,2,3], 0)
generate_node_graph(node.pk)
```
Screenshot 2024-09-17 at 11 34 09

but the dynamic input ports approach gives
```python
import numpy as np

@calcfunctione
def __getitem__(**kwargs):
    idx = kwargs.pop('idx')
    list_of_orms = list(kwargs.values())
    return list_of_orms[idx.value].clone()

inp = {f'list_of_orms_{i}': orm.StructureData(cell=np.random.rand(3,3)) for i in range(3)}
inp['idx'] = orm.Int(0)
node = __getitem__(**inp)
generate_node_graph(node.pk)
```
Screenshot 2024-09-17 at 11 35 36

There is also the new StructureData object that might be json serializable because of the `BaseModel` that resolves this issue (EDIT: no Miki tried). I am also concerned by performance, since we now have to convert a Generator to a list which can be quite time consuming. I feel more and more that we should drop this PR. With the `For` workgraph we can maybe go more into a similar direction as with the while loop and just focus and a nice GUI representation.

superstar54:
The `For` and `While` WorkGraph types are dropped and not used anymore. Instead, the `Map` and `While` context managers are used. Hence, I am closing this PR.

PR #287, opened 2024-09-01
Add aggregate notebook that explains how to aggregate dynamically sized outputs

agoscinski:
Implements suggestions in issue https://github.com/aiidateam/aiida-workgraph/issues/284

~@superstar54 I have an issue with `submit(wait=True)` the output is None (see first example where I kept submit).~ EDIT: I see that this is something related to my local setup and not a general bug.

While I think the section about nested data types at the end is important, it does not really fit into the aggregate notebook. It is also more dealing with aiida-core issues than workgraph. I have no good idea at the moment. I would keep it there and see if we can move it somewhere more meaningful in the future.

PR #288, opened 2024-09-02
Rework parallel doc example using CalcJob's

agoscinski:
The previous example did rely on calcfunctions that are always run sequentially. This example now uses CalcJobs to actually achieve parallel executions.

@superstar54 I removed the old example since it did not make sense to me much, since it was using calcfunctions, but I might be missing something you wanted to actually show with it.


docs/gallery/howto/autogen/parallel.py
```diff
@@ -18,131 +25,82 @@
...
-
-wg.submit(wait=True)
-
+from aiida.calculations.arithmetic.add import ArithmeticAddCalculation
+from aiida.orm import Int, InstalledCode, load_computer, load_code
+from aiida.common.exceptions import NotExistent
+
+# The ArithmeticAddCalculation needs to know where bash is stored
+# For that you need to run in the terminal
+# TODO add verdi command
```
agoscinski:
```suggestion
```

docs/gallery/howto/autogen/parallel.py
```diff
@@ -18,131 +25,82 @@
...
+add_uv = wg.add_task(ArithmeticAddCalculation, x=u, y=v, code=code)
+add_uv.set({"metadata.options.sleep": 5})  # the CalcJob will sleep 5 seconds
+add_xyuv = wg.add_task(
+    ArithmeticAddCalculation,
+    x=add_xy.outputs["sum"],
+    y=add_uv.outputs["sum"],
+    code=code,
+)
+
+wg.run()  # TOOD submit does not workflow
```
agoscinski:
```suggestion
wg.submit(wait=True)
```

docs/gallery/howto/autogen/parallel.py
```diff
@@ -18,131 +25,82 @@
...
 
-X = {"a": Int(1), "b": Int(2), "c": Int(3)}
-y = Int(2)
-z = Int(3)
-wg = WorkGraph("parallel_tasks")
-multiply_parallel_gather1 = wg.add_task(multiply_parallel_gather, X=X, y=y)
-sum1 = wg.add_task(sum, name="sum1")
-# wg.add_link(add1.outputs[0], multiply_parallel_gather1.inputs["uuids"])
-wg.add_link(multiply_parallel_gather1.outputs[0], sum1.inputs[0])
+wg.run()
```
agoscinski:
```suggestion
wg.submit(wait=True)
```

docs/gallery/howto/autogen/parallel.py
```diff
@@ -18,131 +25,82 @@
...
+print("add_uv.ctime", add_uv.ctime)
 
 # %%
-# Generate node graph from the AiiDA process:
-#
-
-from aiida_workgraph.utils import generate_node_graph
-
-generate_node_graph(wg.pk)
+# We can see that both calcjobs are started to run almost at the same time
```
agoscinski:
```suggestion
# We can see that both CalcJob's have been started almost at the same time
```

docs/gallery/howto/autogen/parallel.py
```diff
@@ -18,131 +25,82 @@
...
-#
-
-from aiida_workgraph.utils import generate_node_graph
-
-generate_node_graph(wg.pk)
-
-# %%
-# You can see that the outputs of `multiply_parallel_gather` workgraph is linked to the input of the `sum` task.
-#
+# We can see that the calcfunctions are run with a 5 seconds gap
```
agoscinski:
```suggestion
# We can see that the calcfunctions have been run with a 5 seconds delay
```

superstar54:
Hi @agoscinski , thanks for your efforts on this.

Yes, using `CalcJob` is indeed the correct approach for running jobs in parallel.

The old example, which utilizes a `for` loop to create multiple tasks, reflects many real-world scenarios where parallel execution is essential. For instance:

- Calculating the equation of state, where multiple SCF tasks are created and need to run in parallel.
- Generating multiple surface slabs from a crystal structure.
- Simulating molecule adsorption on the non-equivalent sites of a surface slab.
- Performing XPS calculations on non-equivalent sites.

Besides, in most cases, after these tasks run in parallel, we need to gather and process the results.

The old example is quite valuable for realistic applications, as already mentioned in this [post](https://aiida.discourse.group/t/aiida-workgraph-how-to-handle-dynamic-inputs-and-outputs-of-tasks/457). Therefore, I’d like to suggest keeping the old example but updating it to replace the `calcfunction` with `CalcJob`. Additionally, the provenance graph is very informative and beneficial for users to understand the workflow, so I recommend retaining it as well.

agoscinski:
I wanted to put the for loop back into the notebook but compare it to executing WorkGraphs separately so I can relate it a bit to what is written here https://aiida.discourse.group/t/run-only-one-job-on-local-machine/459/2
but I am not sure in this case how to actually get the time when the process has finished. I first used `mtime` for it as I thought the node is modified after the execution has finished, but the resulting `mtime-ctime` is below the sleeping time. Is there another way to take the time of a process using `submit`? I dont want to time it in the notebook as I need to submit two workgraphs.

```
from aiida.calculations.arithmetic.add import ArithmeticAddCalculation
from aiida.orm import load_code, load_node, load_computer, InstalledCode

load_profile()

from aiida.common.exceptions import NotExistent

# The ArithmeticAddCalculation needs to know where bash is stored
try:
    code = load_code("add@localhost")  # The computer label can also be omitted here
except NotExistent:
    code = InstalledCode(
        computer=load_computer("localhost"),
        filepath_executable="/bin/bash",
        label="add",
        default_calc_job_plugin="core.arithmetic.add",
    ).store()



@task.graph_builder()
def parallel_add(nb_it):
    wg = WorkGraph()
    code = load_code("add@localhost")
    for i in range(nb_it):
        add = wg.add_task(ArithmeticAddCalculation, name=f"add{i}", x=5, y=i, code=code)
        add.set({"metadata.options.sleep": 5})
    return wg
    
wg = WorkGraph(f"parallel_graph_builder")
t = wg.add_task(parallel_add, nb_it=2)

wg.submit(wait=True)
print(load_node(wg.pk).mtime - load_node(wg.pk).ctime)
```

superstar54:
> but the resulting mtime-ctime is below the sleeping time. 

This shouldn't happen. Could you double-check?

agoscinski:
Ah it is because of caching. After I disabled the caching with verdi the caching is still somehow active, not sure why. I will try to reproduce it later and do an issue on aiida-core

docs/source/conf.py
```diff
@@ -71,6 +71,7 @@
     "filename_pattern": "/*",
     "examples_dirs": gallery_src_dirs,  # in sphinx-gallery doc referred as gallery source
     "gallery_dirs": sphinx_src_autogen_dirs,  # path to where to gallery puts generated files
+    "run_stale_examples": True,
```
agoscinski: I tried to use it to enforce rerunning the notebook but it did not help
```suggestion
```

agoscinski:
I had to rename the parallel file so no cached version is used from read the docs. I am still confused why this happened. Since I enforced a rerun on the sphinx-gallery side using `run_stale_examples`, it might be due to some caching of RTD. When the PR has been reviewed, I will change the name of the name after the review.

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,245 @@
...
+wgs = []
+for i in range(2):
+    wg = WorkGraph(f"parallel_wg{i}")
+    wg.add_task(add10_wg, name=f"add10_{i}", integer=i)
+    wgs.append(wg)
+
+# We use wait=False so we can continue submitting
+wgs[0].submit(wait=False)
+# we wait for the last WorkGraph to finish
+wgs[1].submit(wait=True)
```
superstar54:
```suggestion
wgs[0].submit()  # do not wait (by default), so that we can continue to submit next WG.
wgs[1].submit(wait=True)
# we wait for all the WorkGraphs to finish
wgs[0].wait()
```

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,245 @@
...
+# ========================
+# We will parallelize a workgraph by two ways, one time we submit all workgraphs,
+# the other time we use the graph builder to submit once the whole workflow.
+
+
+# This is our initial WorkGraph we want to parallelize
+@task.graph_builder(
+    inputs=[{"name": "integer"}], outputs=[{"name": "sum", "from": "sum_task.result"}]
+)
+def add10_wg(integer):
```
superstar54: Why use this name `add10_wg`?

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,245 @@
...
+wg.submit(wait=True)
+
+
+# %%
+# We look at the times of creation and last change
+print("Time for running with graph builder", add_task.mtime - add_task.ctime)
+
+# %%
+# We can see that the time is more than 5 seconds which means that the two additions
+# were performed in parallel
```
superstar54: This seems to be a wrong description. Could you correct this?

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+# %%
+# Using graph builder
+# -------------------
+
+
+# This graph_builder runs the add10_wg over a loop and its
+@task.graph_builder()
+def parallel_add(nb_it):
```
superstar54: Could you choose a better name for `nb_it`, also and docstring.

superstar54 (changes requested):
Hi @agoscinski thanks for the work. I do have one concern: the current execution time of 2 minutes and 4.786 seconds is quite long.
Here are my suggestions:
- The "Parallelizing WorkGraphs" section is not needed. Take one use case as an example, I have PwRelax workgraph and a large set of structures to relax. Users just need to directly submit multiple workgraphs using a simple loop, without waiting for each to finish.  If users write another workgraph to manage parallel execution, users need to think about how to passing input into the sub-workgraph, and handling potential interruptions of the top-level workgraph. While tracking the provenance of 100 workgraphs submitted together might be a potential benefit, it’s not something most users would require.
- Add a description of the workgraph execution mechanism at the beginning of the notebook: Workgraphs are based on dependency-driven execution, meaning parallel tasks run automatically if the tasks have no dependency on each others, and there are sufficient resources. I recall you demonstrated this well in the EuroScipy tutorial, with a workgraph showing how the process works in parallel (the execution order). You could include that example here, and also show the GUI for better visualization.
- Better to show the GUI for every workgraph: This will help users easily see which tasks will be run in parallel.
- Mention the result-gathering process: Typically, users will want to gather the results after parallel tasks are complete. You can add a link to the aggregate notebook, and if it's not ready yet, you can include a comment that it will be added later.

docs/gallery/howto/autogen/parallel_wf.py
agoscinski: `add10` because it adds 10 to a number. `wg` is I think a bit general indicating the problem that I have when I want to refer a graph builder. It can be a function as defined here, a task when integrated into a WorkGraph, and it is actually a WorkGraph that is executed. Because the whole tutorial is about parallelizing workgraphs I added `wg` into the name, but we can also remove it.

docs/gallery/howto/autogen/parallel_wf.py
agoscinski:
```suggestion
def parallel_add(nb_iterations):
```

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,245 @@
...
+    wg = WorkGraph()
+    for i in range(nb_it):
```
agoscinski:
```suggestion
    for i in range(nb_iterations):
```

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,245 @@
...
+        wg.add_task(add10_wg, name=f"add10_{i}", integer=i)
+    return wg
+
+
+# Submitting a parallel that adds 10 two times to different numbers
+wg = WorkGraph(f"parallel_graph_builder")
+add_task = wg.add_task(parallel_add, name="parallel_add", nb_it=2)
```
agoscinski:
```suggestion
add_task = wg.add_task(parallel_add, name="parallel_add", nb_iterations=2)
```

docs/gallery/howto/autogen/parallel_wf.py
agoscinski:
```suggestion
# We can see that the time is less than 10 seconds which means that the two additions
# were performed in parallel
```

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,245 @@
...
+
+from aiida.engine.daemon.client import get_daemon_client
+
+client = get_daemon_client()
+
+# %%
+# We rerun the last graph builder for 5 iterations
+
+wg = WorkGraph("wg_daemon_worker_1")
+wg.add_task(parallel_add, name="parallel_add", nb_it=5)
```
agoscinski:
```suggestion
wg.add_task(parallel_add, name="parallel_add", nb_iterations=5)
```

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,245 @@
...
+# We increase the number of workers by one. One can also do this in the workgraph GUI.
+
+client = get_daemon_client()
+client.increase_workers(1)
+
+# %%
+# Now we submit again and the time have shortens a bit.
+
+wg = WorkGraph("wg_daemon_worker_2")
+wg.add_task(parallel_add, name="parallel_add", nb_it=5)
```
agoscinski:
```suggestion
wg.add_task(parallel_add, name="parallel_add", nb_iterations=5)
```

agoscinski:
>  The "Parallelizing WorkGraphs" section is not needed. Take one use case as an example, I have PwRelax workgraph and a large set of structures to relax. Users just need to directly submit multiple workgraphs using a simple loop, without waiting for each to finish. If users write another workgraph to manage parallel execution, users need to think about how to passing input into the sub-workgraph, and handling potential interruptions of the top-level workgraph. While tracking the provenance of 100 workgraphs submitted together might be a potential benefit, it’s not something most users would require.

I would not rely on any qe dependent documentation pages as I think we should move them to aiida-tutorials. Also this is more focused on the parallel execution, measuring timings and talking about daemons. If there is a user base for which both use cases might be useful, then we should keep it (at least until more feedback from outside). If it is the running time, then we should work an improving this. I will reduce the sleeps and the iterations at the end to reduce the running time to 1 minute.

> Add a description of the workgraph execution mechanism at the beginning of the notebook: Workgraphs are based on dependency-driven execution, meaning parallel tasks run automatically if the tasks have no dependency on each others, and there are sufficient resources. I recall you demonstrated this well in the EuroScipy tutorial, with a workgraph showing how the process works in parallel (the execution order). You could include that example here, and also show the GUI for better visualization.

This is mentioned in the second sentence of the example. I will include the GUI and write comments in the first example emphasizing on this. Maybe you can note in the code where I should also mention it.

> Better to show the GUI for every workgraph: This will help users easily see which tasks will be run in parallel.

Included!

> Mention the result-gathering process: Typically, users will want to gather the results after parallel tasks are complete. You can add a link to the aggregate notebook, and if it's not ready yet, you can include a comment that it will be added later.

Okay reference it at the end, but the sphinx reference does not work for the moment, since we need to wait for https://github.com/aiidateam/aiida-workgraph/pull/287 to be merged

docs/gallery/howto/autogen/parallel_wf.py
agoscinski: renamed to `add10` I think that is more consistent with the rest

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,276 @@
...
+# -----------------------------------
+# Since each daemon worker can only manage one WorkGraph (handling the results)
+# at a time, one can experience slow downs when running many jobs that can be
+# run in parallel. The optimal number of workers depends highly on the jobs
+# that are run.
+
+from aiida.engine.daemon.client import get_daemon_client
+
+client = get_daemon_client()
+print(f"Number of current daemon workers {client.get_numprocesses()}")
```
agoscinski:
```suggestion
print(f"Number of current daemon workers {client.get_numprocesses()["numprocesses"]}")
```

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,276 @@
...
+
+# %%
+# We rerun the last graph builder with 2 damon workers
+
+client.increase_workers(1)
```
agoscinski:
```suggestion
client.increase_workers(1)
print(f"Number of current daemon workers {client.get_numprocesses()["numprocesses"]}")
```

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,276 @@
...
+# %%
+wg.submit(wait=True)
+print(
+    "Time for running parallelized graph builder with 2 daemons",
+    parallel_add_task_2.mtime - parallel_add_task_2.ctime,
+)
+
+# %%
+# The overhead time has shortens a bit as the handling of the CalcJobs and
+# WorkGraphs could be parallelized.
```
agoscinski:
```suggestion
# The overhead time has shortens a bit as the handling of the CalcJobs and
# WorkGraphs could be parallelized. One can increase the number of iterations
# to see a more significant difference.
```

agoscinski:
I decreased the time now to 1min11s by reusing the graph builder runs from before for the daemon part. Also I just keep 2 iterations as this is enough for the showcase of the effect of daemons. I reduced the sleep time to 3 seconds.

superstar54:
Hi @agoscinski , the docs failed to build. The pre-commit failed. Could you please fix them? I will review them as soon as they pass.

docs/gallery/howto/autogen/parallel_wf.py
```diff
@@ -0,0 +1,278 @@
+"""
```
agoscinski: Note that RTD caches the other file somehow, therefore I renamed it. I don't fully understand this.

docs/gallery/howto/autogen/parallel.py
```diff
@@ -1,148 +1,280 @@
...
+#        verdi config set daemon.worker_process_slots 200
+#        verdi daemon restart
+
+
+# %%
+# Further reading
+# ---------------
+# Now you learned how to run tasks in parallel you might want to know how to
+# aggregate the results of all these parallel tasks (e.g. taking the mean of
+# all computed values). For this you can further read `how to aggregate outputs <aggregate.html>`_`
```
agoscinski:
```suggestion
# all computed values). For this you can further read `how to aggregate outputs <aggregate.html>`_
```

docs/gallery/howto/autogen/parallel.py
```diff
@@ -1,148 +1,280 @@
...
+# %%
+# Reset back to one worker
+client.decrease_workers(1)
 
 # %%
-# You can see that the outputs of `multiply_parallel_gather` workgraph is linked to the input of the `sum` task.
+# Maximum number of active WorkGraphs
+# -----------------------------------
+# Be aware that for the moment AiiDA can only run 200 WorkGraphs at the same time.
+# To increase that limit one can set this variable to a higher value.
```
superstar54, line 280:
```suggestion
# Be aware that for the moment, AiiDA can only run 200 processes (WorkGraph, CalcJob etc) at the same time.
# To increase that limit, one can set this variable to a higher value.
```

docs/gallery/howto/autogen/parallel.py
```diff
@@ -1,148 +1,280 @@
...
+)
 
-from aiida_workgraph import WorkGraph
-from aiida.orm import Int, List
+# %%
+# Increasing number of daemon workers
+# -----------------------------------
+# Since each daemon worker can only manage one WorkGraph (handling the results)
+# at a time, one can experience slow downs when running many jobs that can be
+# run in parallel. The optimal number of workers depends highly on the jobs
```
superstar54, line 222:
```suggestion
# One can experience slow downs when running many jobs (e.g., 100 jobs) that can be
# run in parallel. The optimal number of workers depends highly on the jobs
```

docs/gallery/howto/autogen/parallel.py
```diff
@@ -1,148 +1,280 @@
...
+print(
+    "Time for running parallelized graph builder with 2 daemons",
+    parallel_add_task_2.mtime - parallel_add_task_2.ctime,
+)
 
 # %%
-# Generate node graph from the AiiDA process:
-#
+# The overhead time has shortens a bit as the handling of the CalcJobs and
+# WorkGraphs could be parallelized. One can increase the number of iterations
```
superstar54: Are you sure on `he overhead time has shortens a bit`?

Looking at the time, there is no improvement.
```
Time for running parallelized graph builder 0:00:11.262496
Time for running parallelized graph builder with 2 daemons 0:00:11.264958
```

docs/gallery/howto/autogen/parallel.py
```diff
@@ -1,148 +1,280 @@
...
+# Maximum number of active WorkGraphs
+# -----------------------------------
+# Be aware that for the moment AiiDA can only run 200 WorkGraphs at the same time.
+# To increase that limit one can set this variable to a higher value.
+#
+#     .. code-block:: bash
 #
+#        verdi config set daemon.worker_process_slots 200
+#        verdi daemon restart
+
```
superstar54: It would be good to add a link to the [performance page](https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/installation.html#tuning-performance).

superstar54 (changes requested):

superstar54:
The `parallel` doc is refactored in another [PR](https://github.com/aiidateam/aiida-workgraph/pull/517), hence I am closing this one.

PR #291, opened 2024-09-05
Create a short version of the workgraph data for the connectivity analysis

### LINKED ISSUE #290: Setting computer in ShellJob raises pickle error
the code
```python
from aiida import orm

wg = WorkGraph()
task = wg.add_task(
    "ShellJob",
    name="ls_task",
    command="ls",
)
task.set({"metadata.computer": orm.load_computer('localhost')})
wg.run()
```
raises the error
```
File ~/micromamba/envs/autosubmit-dev/lib/python3.11/site-packages/aiida/orm/entities.py:241, in Entity.__getstate__(self)
    239 def __getstate__(self):
    240     """Prevent an ORM entity instance from being pickled."""
--> 241     raise InvalidOperation('pickling of AiiDA ORM instances is not supported.')

InvalidOperation: pickling of AiiDA ORM instances is not supported.
```
The whole log 
[errorlog.txt](https://github.com/user-attachments/files/16886815/errorlog.txt)

superstar54:
When analyzing the connectivity of the graph, The actual input data of the tasks is not needed. Therefore, the `ConnectivityAnalysis` class from `node_graph` tries to deepcopy to create a short version. However, it raised an error when deepcoping of the orm.computer, as reported in #290 .

~~In principle, we can fix this in `ConnectivityAnalysis` in `node_graph`, but we can also fix it in workgraph. This PR tries to create the short version data before passing to `ConnectivityAnalysis`, and avoids deepcopy the data of the tasks but only picks the info that is needed.~~
This is fixed in the `node-graph` package, and this PR bumped `node-graph` to the latest version.

### DIFF
--- a/aiida_workgraph/utils/analysis.py
+++ b/aiida_workgraph/utils/analysis.py
@@ -214,7 +214,6 @@ class WorkGraphSaver:
                     prop["value"] = PickledLocalFunction(prop["value"]).store()
             self.wgdata["tasks"][name] = serialize(task)
         # nodes is a copy of tasks, so we need to pop it out
-        self.wgdata.pop("nodes")
         self.wgdata["error_handlers"] = serialize(self.wgdata["error_handlers"])
         self.wgdata["context"] = serialize(self.wgdata["context"])
         self.process.base.extras.set("_workgraph", self.wgdata)
@@ -277,8 +276,6 @@ class WorkGraphSaver:
             return
         for name, task in wgdata["tasks"].items():
             wgdata["tasks"][name] = deserialize_unsafe(task)
-        # also make a alias for nodes
-        wgdata["nodes"] = wgdata["tasks"]
         wgdata["error_handlers"] = deserialize_unsafe(wgdata["error_handlers"])
         return wgdata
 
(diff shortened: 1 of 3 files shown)


aiida_workgraph/utils/analysis.py
```diff
@@ -294,6 +291,9 @@ def check_diff(
         from node_graph.analysis import DifferenceAnalysis
 
         wg1 = self.get_wgdata_from_db(restart_process)
+        # also make a alias for nodes
+        wg1["nodes"] = wg1["tasks"]
+        self.wgdata["nodes"] = self.wgdata["tasks"]
```
agoscinski: why do you not pop the item here as in the code changes below?

agoscinski (approved):
Thanks fixes it for me.

aiida_workgraph/utils/analysis.py
superstar54: Thanks! Keeping both `tasks` and `nodes` will not raise an error in `DifferenceAnalysis`. But for consistency, I changed it to be the same as the code below.

GeigerJ2:
@superstar54 you are too quick :D was about to report that I had the same error today, and this also fixes it for me!

PR #294, opened 2024-09-06
Use type_mapping from aiida-workgraph

superstar54:
`aiida-workgraph` borrows some socket type from the `node-graph` package, which make it hard to control the property of the socket.

This PR removes all socket types from `node-graph`.


aiida_workgraph/properties/builtins.py
```diff
@@ -1,11 +1,91 @@
...
+        super().__init__(name, description, default, update)
+
+    def set_value(self, value):
+        # run the callback function
+        if isinstance(value, (int, type(None))):
+            self._value = value
+            if self.update is not None:
+                self.update()
+        else:
+            raise Exception("{} is not a integer.".format(value))
```
GeigerJ2, line 24: Isn't that a bit of a mix between an f-string and `.format`?

aiida_workgraph/properties/builtins.py
```diff
@@ -1,11 +1,91 @@
...
+
+
+class PropertyFloat(TaskProperty, SerializeJson):
+    """A new class for float type."""
+
+    identifier: str = "workgraph.float"
+    data_type = "Float"
```
GeigerJ2, line 31: Is this uppercase because it denotes an `orm.Float`?

GeigerJ2 (approved):
Looks all reasonable to me. Worked also for adding the `node_graph` property, getting the WorkGraph `Socket`s, as expected, not the node-graph `Socket`s.

Why the bump to `node-graph` v0.0.17? Is this just a side thing, or does it have deeper meaning?

I'd say LGTM!

aiida_workgraph/properties/builtins.py
superstar54: No special meaning, the `data_type` is case-insensitive, so it doesn't matter.

superstar54:
> Why the bump to node-graph v0.0.17? Is this just a side thing, or does it have deeper meaning?

I updated the `node-graph` to support passing `type_mapping` as an argument when creating the task. So this is needed.

PR #295, opened 2024-09-06
GUI: add task control for int, float, string and structure

### LINKED ISSUE #296: GUI: show node properties (detail view)
## Background: 
Currently, the aiida-workgraph GUI displays a simplified view of node sockets, showing only the socket names. While this minimalistic view helps maintain a clean interface, it limits the user's ability to quickly grasp detailed information about the node details, such as data types and values passed between nodes.

## Objective
Implement a feature that allows users to toggle between a "Short View" and a "Detailed View" of node sockets within the aiida-workgraph GUI.

## Short view
The default view that displays only the socket names. This view is suitable for a simplified and clean overview of the workflow.

image

## Detail view
An expanded view that shows additional information about each socket, including the visualization of the data. As shown in PR #295 

image

## Custom Controls

For each socket type, developers can implement a `Control` component that dictates how the data for that socket is visualized in the "Detailed View". This approach provides flexibility and can accommodate custom representations, but we need to figure out how to add the `Control` as a plugin.

## Viewer task
Add a built-in `viewer` task on the frontend, so that one can visualize the output sockets.

superstar54:
See issue #296 

Here is an example that shows the value of int, float, string and StructureData sockets:

```python
from ase.build import molecule

h2o = molecule("H2O")
h2o.pbc = [True, True, True]
h2o.center(vacuum=2.0)

structure = orm.StructureData(ase=h2o)

@task.calcfunction()
def add(x: int, y: float, z: str, structure: orm.StructureData):
    return x + y


wg = WorkGraph("add_multiply_workflow")
add_task = wg.add_task(add, name="add1", x=1, y=2.1, z="abc", structure=structure)
wg.add_task(multiply, name="multiply1", x=add_task.outputs["result"], y=3)
wg.run()
```
Then open the web app, and visualize the WorkGraph, here is a screenshot from the web app:

image

## TODO
How to handle serializer registered in PythonJob?


agoscinski:
Can this be also turned off? I am just a bit worried if people have huge structures that the GUI becomes quite slow. Not sure if this a valid worry as I don't know much about the computational cost for rendering atomic structures

superstar54:
> Can this be also turned off?

Yes, there is a toggle button so that the user can switch between `short view` and `node detail view.`

PR #325, opened 2024-09-19
Removing QE dependencies and tutorials

agoscinski:
This is for testing the running time on RTD without qe to see if it is really such a significant factor.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -17,7 +17,6 @@ build:
       - verdi presto
       - verdi daemon start
       - verdi status
-      - aiida-pseudo install sssp -x PBEsol
       - verdi group list
       - cat /proc/cpuinfo | grep processor | wc -l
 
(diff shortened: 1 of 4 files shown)


agoscinski:
Drastic reduction 576s -> 344s

superstar54:
Thanks for the test, then we can revert them to normal notebooks, and run them locally. We also create another section, or use the `Gallery` section to hold all the real-world use cases.

agoscinski:
> Thanks for the test, then we can revert them to normal notebooks, and run them locally. We also create another section, or use the Gallery section to hold all the real-world use cases.

For a gallery we would need to refactor them. They are a bit convoluted with introducing concepts and presenting use cases. Probably it would be better if they link to the corresponding howtos/concepts. If feel like `zero to hero` does not contain anything which is not already covered in the doc. I would try to merge it with quick start. So I only see the two use cases `N2 energy` computation and `EOS`. There is also the `Use protocol to set input parameters (Experimental)` part that I cannot pin down where to put.

superstar54:
Superseded by #609

PR #326, opened 2024-09-19
Update parser signature for `aiida-shell` v0.8, extend ShellJob docs, use `prepare_shell_job_inputs`

GeigerJ2:
Still needs some work, but already creating the PR now to see if CI runs through.

### DIFF
--- a/aiida_workgraph/engine/utils.py
+++ b/aiida_workgraph/engine/utils.py
@@ -131,34 +131,22 @@ def prepare_for_python_task(task: dict, kwargs: dict, var_kwargs: dict) -> dict:
 
 def prepare_for_shell_task(task: dict, kwargs: dict) -> dict:
     """Prepare the inputs for ShellJob"""
-    from aiida_shell.launch import prepare_code, convert_nodes_single_file_data
-    from aiida.common import lang
-    from aiida.orm import AbstractCode
+    from aiida_shell.launch import (
+        convert_nodes_single_file_data,
+        prepare_shell_job_inputs,
+    )
 
-    command = kwargs.pop("command", None)
-    resolve_command = kwargs.pop("resolve_command", False)
     metadata = kwargs.pop("metadata", {})
-    # setup code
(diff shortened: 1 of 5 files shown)


aiida_workgraph/engine/utils.py
```diff
@@ -152,11 +153,19 @@ def prepare_for_shell_task(task: dict, kwargs: dict) -> dict:
...
             nodes[key[6:]] = kwargs.pop(key)
     metadata.update({"call_link_label": task["name"]})
+
+    arguments = kwargs.pop("arguments", [])
+    # Split on whitespace if arguments given as `str`
+    if isinstance(arguments, str):
+        arguments = shlex.split(arguments)
+    else:
+        lang.type_check(arguments, list, allow_none=True)
+
```
superstar54: Hi @GeigerJ2 , thanks for the work, the idea is to call the new `prepare_shell_job_inputs` (from aida-shell) inside this `prepare_for_shell_task` function, to replace lines 142-150 (prepare code etc), 157-162 (slit argument). Thus we don't need to manually update this function if aiida-shell updates.

GeigerJ2: Ah, OK, I see. Yeah, I was also wondering why we duplicate some code from that function here, but thought there might be a reason to handle some things differently in WorkGraph.

GeigerJ2: Wasn't aware of the relevant [PR](sphuber/aiida-shell#103) in `aiida-shell`. Now using the `prepare_shell_job_inputs` helper function, and just updating the `kwargs` `nodes` and `metadata` as required.

GeigerJ2:
Finally, CI passing 😅 OK, this should be ready for review, @superstar54 :)

superstar54:
Hi @GeigerJ2 , thanks for the work! I modified the `prepare_for_shell_task` function because some of the code is already covered by the `prepare_shell_job_inputs`.

PR #331, opened 2024-09-20
add padding for scope

superstar54:
Defines custom padding for a scope layout. The padding values are used by the ScopesPlugin to avoid node overlapping with the socket of the parent node.
Before, the `conditions` socket overlaps with task:

Screenshot from 2024-09-20 10-01-05

After:
Screenshot from 2024-09-20 09-56-02

Note: we need to update three places: 1) web app, 2) widget 3) HTML template. This is not good; we need to refactor the code in the future so that we create a library for all.


docs/source/howto/html/if_task.html
```diff
@@ -61,7 +61,7 @@
...
...}, {"name": "x"}], "properties": {"metadata": null, "metadata.store_provenance": null, "metadata.description": null, "metadata.label": null, "metadata.call_link_label": null, "metadata.disable_cache": null, "x": null, "y": null, "_wait": null}, "outputs": [], "position": [240, 240], "children": []}}, "links": [{"from_socket": "result", "from_node": "add1", "to_socket": "x", "to_node": "add2", "state": false}, {"from_socket": "result", "from_node": "condition1", "to_socket": "conditions", "to_node": "if_true", "state": false}, {"from_socket": "result", "from_node": "add1", "to_socket": "x", "to_node": "multiply1", "state": false}, {"from_socket": "result", "from_node": "condition1", "to_socket": "conditions", "to_node": "if_false", "state": false}, {"from_socket": "result", "from_node": "add2", "to_socket": "true", "to_node": "select1", "state": false}, {"from_socket": "result", "from_node": "multiply1", "to_socket": "false", "to_node": "select1", "state": false}, {"from_socket": "result", "from_node": "condition1", "to_socket": "condition", "to_node": "select1", "state": false}, {"from_socket": "result", "from_node": "select1", "to_socket": "x", "to_node": "add3", "state": false}]}
```
agoscinski, line 64: what has here changed? it is a bit hard to detect^^ Or maybe more meaningful, how do you generate this code? Could you make comment in the code explaining this?

aiida_workgraph/web/frontend/src/rete/default.ts
```diff
@@ -125,14 +125,21 @@ function createDynamicNode(nodeData: any) {
   return node;
 }
 
+const customPadding = () => ({
```
agoscinski, line 128: Can you maybe give it a bit more meaningful name or explain it a bit? I guess this is the padding relative to the parent node?

agoscinski (changes requested):
Looks overall good, just a bit doc missing.

> Note: we need to update three places: 1) web app, 2) widget 3) HTML template. This is not good; we need to refactor the code in the future so that we create a library for all.

Can you make a note in some issue referencing this PR or open a new issue to just keep track of it?

docs/source/howto/html/if_task.html
superstar54: You can think `workgraphData = get_short_versioin(wg.to_dict())`, so it contains all the information on the graph and tasks, inputs, outputs, links etc.

Most of the time, only the uuid changes if we use the same workgraph version. However, because I ran it with new WorkGraph version, so the data is also changed.

aiida_workgraph/web/frontend/src/rete/default.ts
superstar54: Yes, I added in https://github.com/aiidateam/aiida-workgraph/pull/331/commits/109c9cf8f62c48c48088ca40d45fab78ab7e27cf

superstar54:
> Can you make a note in some issue referencing this PR or open a new issue to just keep track of it?

Yes, I created an [issue](https://github.com/aiidateam/aiida-workgraph-web-ui/issues/12)

agoscinski (approved):

PR #334, opened 2024-09-20
Add doc howto: Recursive workflows with Graph Builder

agoscinski:
The new doc how-to explains how to create a recursive workflow with the graph builder.

@superstar54 could you explain the mechanic why we need to import the function from a module?

I started to add like a short section at the start of the how-to referring to concepts the user should first familiarize him/herself.


superstar54:
This doc is outdated, and a similar recursive workflow is implemented using the `Graph` task, hence I am closing this PR.

PR #336, opened 2024-09-22
Fix recursive workgraph

### LINKED ISSUE #333: Recursive WorkGraph with Graph Builder can only be run for the amount of redefinition of it
Because @giovannipizzi mentioned that for a project that need to use recursive workflows I wanted to do an example for recursive workflows. It kind of works, but there are some problems with referring to the the graph builder itself

```python
from aiida_workgraph import task, WorkGraph
load_profile()

@task.calcfunction
def identity(x):
    return x.clone()

@task.calcfunction
def multiply(x, y):
    return x*y

@task.graph_builder(outputs = [{"name": "result", "from": "multiply.result"}])
def recursive(n):
    wg = WorkGraph()
    if n == 1:
        mytask = wg.add_task(identity, name="multiply", x=n)
        return wg 
    t = wg.add_task(recursive, name="recursive", n=n-1)
    wg.add_task(multiply, name="multiply", x=n, y=t.outputs["result"])
    return wg
```
I have to rerun the `recursive` definition the amount of time I want to use it in the recursion
```python
wg = WorkGraph()
mytask = wg.add_task(recursive, n=3) # I have to run the above cell defining recursive 3 times so this works
wg.run()
```

superstar54:
Fix #333 .

There seems to be a bug in `cloudpickle`, that the decorated function is not in the `globals`.
This PR provides a temporary solution.

There is also a bug related to `group_outputs`, which is fixed in #335 .

### DIFF
--- a/aiida_workgraph/engine/workgraph.py
+++ b/aiida_workgraph/engine/workgraph.py
@@ -1032,6 +1032,13 @@ class WorkGraphEngine(Process, metaclass=Protect):
 
             self.report(f"Run task: {name}, type: {task['metadata']['node_type']}")
             executor, _ = get_executor(task["executor"])
+            # Add the executor to the globals so that it can be used in the task
+            # in the case of recursive workgraph
+            # We also need to rebuild the Task calss and attach it to the executor
+            if task["metadata"]["node_type"].upper() == "GRAPH_BUILDER":
+                task_class = Task.from_dict(self.ctx._tasks[name])
+                executor.node = executor.task = task_class.__class__
+                executor.__globals__[executor.__name__] = executor
             # print("executor: ", executor)
             args, kwargs, var_args, var_kwargs, args_dict = self.get_inputs(name)
             for i, key in enumerate(self.ctx._tasks[name]["args"]):
--- a/aiida_workgraph/decorator.py
+++ b/aiida_workgraph/decorator.py
@@ -455,9 +455,9 @@ def build_task_from_workgraph(wg: any) -> Task:
         "type": tdata["metadata"]["task_type"],
         "is_pickle": False,
     }
+    tdata["metadata"]["group_outputs"] = group_outputs
     tdata["executor"] = executor
     task = create_task(tdata)
-    task.group_outputs = group_outputs
     return task
 
 
@@ -501,6 +501,8 @@ def generate_tdata(
     properties: List[Tuple[str, str]],
     catalog: str,
     task_type: str,
+    group_inputs: List[Tuple[str, str]] = None,
+    group_outputs: List[Tuple[str, str]] = None,
     additional_data: Optional[Dict[str, Any]] = None,
 ) -> Dict[str, Any]:
     """Generate task data for creating a task."""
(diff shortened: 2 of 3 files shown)


aiida_workgraph/engine/workgraph.py
```diff
@@ -1032,6 +1032,13 @@ def run_tasks(self, names: t.List[str], continue_workgraph: bool = True) -> None
 
             self.report(f"Run task: {name}, type: {task['metadata']['node_type']}")
             executor, _ = get_executor(task["executor"])
+            # Add the executor to the globals so that it can be used in the task
+            # in the case of recursive workgraph
+            # We also need to rebuild the Task calss and attach it to the executor
```
agoscinski, line 1037:
```suggestion
            # We also need to rebuild the Task calls and attach it to the executor
```

aiida_workgraph/engine/workgraph.py
```diff
@@ -1032,6 +1032,13 @@ def run_tasks(self, names: t.List[str], continue_workgraph: bool = True) -> None
 
             self.report(f"Run task: {name}, type: {task['metadata']['node_type']}")
             executor, _ = get_executor(task["executor"])
```
agoscinski, line 1034: What kind of task is this? When I create add a task in WorkGraph there is no `task["executor"]` but I can do `task.get_executor()`

aiida_workgraph/engine/workgraph.py
```diff
@@ -1032,6 +1032,13 @@ def run_tasks(self, names: t.List[str], continue_workgraph: bool = True) -> None
...
+            # Add the executor to the globals so that it can be used in the task
+            # in the case of recursive workgraph
+            # We also need to rebuild the Task calss and attach it to the executor
+            if task["metadata"]["node_type"].upper() == "GRAPH_BUILDER":
+                task_class = Task.from_dict(self.ctx._tasks[name])
+                executor.node = executor.task = task_class.__class__
+                executor.__globals__[executor.__name__] = executor
```
agoscinski, line 1041: My concern is that executors will have the same name as it is not always unique and then override each other. For calcfunctions for example it is the name of the function which is reasonable, as it is only overridden if a function with the same name is defined. But for the case of local functions, this overrides any global definition.

```python
from aiida_workgraph import task, WorkGraph
from aiida.engine import calcfunction

load_profile()

@task.graph_builder()
def my_add():
    @calcfunction
    def add(x, y):
        return x+y
    wg = WorkGraph()
    task = wg.add_task(add, x=1, y=1)
    return wg
    
wg = my_add()
print(wg.tasks["add1"].get_executor()['name']) # out 'add' but better 'my_add.add'
```

I guess solving this issue requires much more work and since we don't have any examples defining calcfunctions locally (however we have to load codes locally in graph_builder), I think it is not very crucial, but an issue would be nice to keep this in mind.

agoscinski (changes requested):

superstar54:
Replaced by https://github.com/aiidateam/aiida-workgraph/pull/587

PR #343, opened 2024-11-18
Pass list of names for task input/output directly

GeigerJ2:
So far, one would always have to pass a list of dictionaries, specifying at least the `"name"` identifier to a `Task`s inputs/outputs, such as:

```python
SeekpathTask = build_task(
    seekpath_structure_analysis,
    outputs=[
        {"name": "primitive_structure"},
        {"name": "explicit_kpoints"},
    ],
)
```

With this PR, it is now possible to just directly specify a list of strings, which will internally be automatically converted to a list of dictionaries, with the `"name"` key assigned to it. With this, it is now possible to write the above function call shorter, as:

```python
SeekpathTask = build_task(
    seekpath_structure_analysis,
    outputs=["primitive_structure", "explicit_kpoints"]
)
```

To avoid confusion, I only allow _either_ a list of `str`, **or** a list of `dict`, but not of mixed types. I added tests for the `validate_task_inout` function, as well as one test for the full call to `build_graph`.

### DIFF
--- a/aiida_workgraph/decorator.py
+++ b/aiida_workgraph/decorator.py
@@ -6,7 +6,7 @@ from aiida.orm.nodes.process.calculation.calcfunction import CalcFunctionNode
 from aiida.orm.nodes.process.workflow.workfunction import WorkFunctionNode
 from aiida.engine.processes.ports import PortNamespace
 from aiida_workgraph.task import Task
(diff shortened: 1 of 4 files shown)


superstar54 (approved):
@GeigerJ2 , thanks for the nice work! LGTM!

PR #346, opened 2024-11-19
Hide metadata inputs for tasks in HTML visualization.

### LINKED ISSUE #119: Hide all unnecessary `metadata` input sockets from `Task` visualization
Probably short first hack could be already `startswith(metadata)` :D

- `metadata` only shown when decorated with `@task.calcfunction()`

GeigerJ2:
Fixes #119.


aiida_workgraph/widget/src/widget/__init__.py
```diff
@@ -46,6 +46,14 @@ def from_node(self, node: Any) -> None:
         tdata["label"] = tdata["identifier"]
         for input in tdata["inputs"].values():
             input.pop("property")
+
+        if not show_metadata:
+            inputs = {}
+            for input_k, input_v in tdata["inputs"].items():
+                if not input_k.startswith("metadata"):
```
superstar54: I think we still show the top-level `metadata` port, but not expand it.
```suggestion
                if not input_k.startswith("metadata."):
```

superstar54 (changes requested):
Hi @GeigerJ2 , instead of handling `metadata` specially, it would be good to also handle other namespace ports. I suggest to use a parameter to control it.
For example, we can add an attribute to the node, e.g., `show_socket_level`, and the default value is `0` to only the top level, if it is `1`, then show all `root.sub`, and so on.

GeigerJ2:
> Hi @GeigerJ2 , instead of handling `metadata` specially, it would be good to also handle other namespace ports. I suggest to use a parameter to control it. For example, we can add an attribute to the node, e.g., `show_socket_level`, and the default value is `0` to only the top level, if it is `1`, then show all `root.sub`, and so on.

Nice idea, I like it! Will implement it!

GeigerJ2:
Hi, @superstar54, I generalized it in my latest commit. The namespace nesting depth for the shown sockets can now be controlled by a parameter of the `to_html` method of a `Task`. I assumed this nesting only applies to top-level keys, and I'm not sure if this covers all use cases. So please let me know if something else comes to mind here. I'll add tests then afterwards.

superstar54:
Thanks for the update!
apart from the HTML, it also good to add `show_socket_depth` as a attribute of `Task`, and then use it in the widget, as follow:

```diff
def _repr_mimebundle_(self, *args: Any, **kwargs: Any) -> any:

        if self._widget is None:
            print(WIDGET_INSTALLATION_MESSAGE)
            return
        # if ipywdigets > 8.0.0, use _repr_mimebundle_ instead of _ipython_display_
+        self._widget.from_node(self, show_socket_depth=self.show_socket_depth)
-        self._widget.from_node(self)
        if hasattr(self._widget, "_repr_mimebundle_"):
            return self._widget._repr_mimebundle_(*args, **kwargs)
        else:
            return self._widget._ipython_display_(*args, **kwargs)
```
This way, the user can control the widget.

superstar54 (approved):
LGTM!

PR #347, opened 2024-11-19
Visualize `nodes` inputs for ShellJob tasks

### LINKED ISSUE #117: Add socket visualization for AiiDA `Data` nodes to `WorkGraph` display
To give some background on my particular use case: I'm using `WorkGraph` to create a workflow based on various `ShellJob`s with input/output files being passed around between the `Task`s.

When visualizing the `WorkGraph` I would like to see the input sockets of *absolute* file/folder-based `Data` nodes[^1], that is, they don't originate from other tasks of the `WorkGraph`. Possibly, they could be shown with a different styling than the `Task` nodes. Adding this information would allow one to trace back the full flow of (filed/folder-based) data in the `WorkGraph`.

To provide an example — with this snippet[^2]:

```python
from aiida.orm import SinglefileData

wg = WorkGraph("ShellJob")

wg_task = wg.tasks.new(
    "ShellTask",
    name='shelltask',
    command='cat',
    arguments='{file_a} {file_b}',
    nodes={
        'file_a': SinglefileData.from_string('string a'),
        'file_b': SinglefileData.from_string('string b'),
    }
)
```

Visualizing `wg` (the full `WorkGraph`) gives:

i.e. only the `Code` input is shown. However, for `wg_task` the information is excessive (warrants separate PR):

Among those, one can see the two `nodes.file_[a,b]` inputs that I'm interested in:

So I propose to add those also to the full `WorkGraph` visualization.

[^1]: Though, it might also be interesting to visualize all nodes that derive from `Data`. I'm not sure if this is already being done right now? Of what types are all the `metadata` options? AiiDA types or Python base types?
[^2]: Issue regarding the naming here #118

GeigerJ2:
Fixes #117.

Possibly visualize more inputs or other data?

### DIFF
--- a/aiida_workgraph/utils/__init__.py
+++ b/aiida_workgraph/utils/__init__.py
@@ -732,7 +732,9 @@ def workgraph_to_short_json(
             {"name": name, "identifier": input["identifier"]}
             for name, input in task["inputs"].items()
             if name in task["args"]
+            or (task["identifier"].upper() == "SHELLJOB" and name.startswith("nodes."))
         ]
+
         properties = process_properties(task)
         wgdata_short["nodes"][name] = {
             "label": task["name"],


superstar54 (approved):
LGTM!

PR #349, opened 2024-11-19
Make `"name"` specifier of `parser_outputs` optional

### LINKED ISSUE #167: Make `"name"` specifier of `parser_outputs` optional
From the same discussion as #166 with @mbercx and @superstar54, also came this issue: I'm getting quite confused these days with all the string annotations one needs to correctly set up. They seem a bit cumbersome and repetetive, e.g., having to specify:
```python
parser_outputs=[{"name": "energy"}, {"name": "structure"}, {"name": "forces"}]
```
for using a custom parser when running a `ShellJob`. In this case, I propose we make the `"name"` key optional, and allow users to just specify a list (or, make the argument fully optional and inspect from the parser actual outputs of the parser function, though, that might be just wishful thinking, as it's passed as PickledData). I understand that other options could be specified here, e.g. data type, so we need to find a balance between extensibility and ease of use, so the default could be `"name"`, while other/multiple identifiers can still be used.

There are more places where one has to add such string identifiers, the `"General"` comes to mind, e.g., from here:
```python
SeekpathTask = build_task(
    seekpath_structure_analysis,
    outputs=[["General", "primitive_structure"], ["General", "explicit_kpoints"]],
)
```
where `"General"` could also be the default and users could just provide a simple list, as I think came already up on our first `WorkGraph` coding day

GeigerJ2:
Fixes #167.

Using the same `validate_task_inout` function as for task `inputs` and `outputs`.

### DIFF
--- a/aiida_workgraph/decorator.py
+++ b/aiida_workgraph/decorator.py
@@ -379,6 +379,8 @@ def build_shelljob_task(
         tdata["outputs"][output["name"]] = output
     outputs = [] if outputs is None else outputs
     parser_outputs = [] if parser_outputs is None else parser_outputs
+    parser_outputs = validate_task_inout(parser_outputs, "parser_outputs")
+
     outputs = [
         {"identifier": "workgraph.any", "name": ShellParser.format_link_label(output)}
         for output in outputs


superstar54 (approved):
LGTM!

PR #350, opened 2024-11-19
Update root property in the case of nest namespace

superstar54:
In #341 , we set the default value of the namespace to None, which cause the function to update nest dict failed.

This PR fixes it, and also add a test for the `set` method.


GeigerJ2 (approved):
LGTM!

PR #351, opened 2024-11-19
Avoid passing `None` to arguments of `prepare_shell_job_inputs`

### LINKED ISSUE #348: Capture `ValueError` from `aiida-shell` for non-existing executable
The `prepare_shell_job_inputs` function from `aiida-shell` checks if it can resolve the path to a given executable. If not, it raises a `ValueError` to indicate that the executable is not available. However, if one tries to run a `ShellJob` `Task` via the WorkGraph, this error is not captured. Instead, the task is tried to run normally, and eventually fails (as can be seen from `verdi process report` on the failed task).

GeigerJ2:
Originally noted in #348, and fixes #348.

The reason for the bug was that in this line:
https://github.com/aiidateam/aiida-workgraph/blob/8f6260b44595f79471f6ea70e8efe0f0e158cdab/aiida_workgraph/engine/utils.py#L139
iteration was done over all input arguments of the `prepare_shell_job_inputs` function, using `None` as the default for the ones that weren't contained in the `inputs` dictionary passed to the `prepare_for_shell_task` function of WorkGraph. This would lead to `None` values being passed explicitly to the arguments of `prepare_shell_job_inputs`, in particular `resolve_command`, effectively overriding its default `True` value.

Inside `aiida-shell` the check for the absolute path of the executable inside the `prepare_code` function (see [here](https://github.com/sphuber/aiida-shell/blob/14866d1450aa252ec414373e86e98611e2eae9db/src/aiida_shell/launch.py#L176-L184)) would thus be skipped, and the WorkGraph would just fail silently at a later stage. With this PR, only the input arguments that are also contained in the `inputs` of the `prepare_for_shell_task` are being explicitly passed, while the rest is left unset.

### DIFF
--- a/aiida_workgraph/decorator.py
+++ b/aiida_workgraph/decorator.py
@@ -379,6 +379,8 @@ def build_shelljob_task(
         tdata["outputs"][output["name"]] = output
     outputs = [] if outputs is None else outputs
     parser_outputs = [] if parser_outputs is None else parser_outputs
+    parser_outputs = validate_task_inout(parser_outputs, "parser_outputs")
+
     outputs = [
         {"identifier": "workgraph.any", "name": ShellParser.format_link_label(output)}
         for output in outputs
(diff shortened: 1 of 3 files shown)


superstar54 (approved):
LGTM! Please merge if all tests pass.

PR #374, opened 2024-12-03
Move web and widget to the  root of the package

superstar54:


superstar54:
Repalced by #376 and #375

PR #387, opened 2024-12-09
Refactor `AIIDA_CONFIG_FOLDER` -> `AiiDAConfigDir`

khsrali:
After https://github.com/aiidateam/aiida-core/pull/6610 by @unkcpz  `AIIDA_CONFIG_FOLDER` has changed to `AiiDAConfigDir`, 
This PR refactors that accordingly. 
(a similar PR opened on https://github.com/aiidateam/aiida-pythonjob/pull/9)

### DIFF
--- a/src/aiida_workgraph/config.py
+++ b/src/aiida_workgraph/config.py
@@ -1,5 +1,6 @@
 import json
 from aiida.manage import get_config
+from pathlib import Path
 
 WORKGRAPH_EXTRA_KEY = "_workgraph"
 WORKGRAPH_SHORT_EXTRA_KEY = "_workgraph_short"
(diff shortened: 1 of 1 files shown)


src/aiida_workgraph/config.py
```diff
@@ -11,7 +11,7 @@
 
 def load_config() -> dict:
     """Load the configuration from the config file."""
-    config_file_path = AIIDA_CONFIG_FOLDER / "workgraph.json"
+    config_file_path = AiiDAConfigDir.get() / "workgraph.json"
```
unkcpz: Try public API maybe? as mentioned in https://github.com/aiidateam/aiida-core/issues/6656#issuecomment-2531063738

BTW, is the workgraph.json config bind to a config or it is better to bind to a profile? @superstar54

superstar54:
Hi @khsrali , thanks for taking care of this issue! Could you use the API suggested by @unkcpz?

src/aiida_workgraph/config.py
superstar54: Thanks for pointing out this. For the moment, it's a global setting for all. But maybe in the future, we could also support overriding it for each profile.

superstar54 (approved):
LGTM! Thanks!

PR #394, opened 2025-01-17
Small fix in the doc

mikibonacci:
changing a variable name to avoid the override of the task object used to decorate the python functions.


superstar54 (approved):
LGTM, thanks for the fix!

PR #396, opened 2025-01-28
Fix PythonJob need `aiida_workgraph` on remote computer

superstar54:
In the decorated function of PythonJob, we attached the WorkGraph's `Task` class to the function, this will be pickled together with the function, thus, when unpickling the function on the remote computer, it require `aiida-workgraph`. 

Thsi PR deletes the workgraph related attributes of the decorated func if exist.


mikibonacci (approved):
Hi @superstar54 , I tested it and it works, LGTM! Thanks a lot for the fix!

PR #410, opened 2025-02-24
Migrate workgraph data from `base.extras` to `base.attributes`

### LINKED ISSUE #401: use _updatable_attributes instead of `base.extras`
```python
@classproperty
    def _updatable_attributes(cls) -> Tuple[str, ...]:  # noqa: N805
```

### LINKED ISSUE #402: Separate state/metadata in ctx

superstar54:
# Engine

## Inputs
In WorkGraph engine, the `workgraph data`, including the tasks, links, inputs and outputs of the tasks etc, are passed as inputs of the WorkGraphEngine.

This `workgraph data` will be used in:
- execution of the WorkGraph
- re-load the WorkGraph from the database

In the old design, the `workgraph data` is stored in the `base.extras` of the node. However, there are several drawbacks:
- If the `base.extas` is deleted, one can not restore the WorkGraph
- could have a security issue if the data is modifed by other people.

In this PR, we save the `workgraph data` as an attribute of the node (`WorkGraphNode`), when creating the process (`WorkGraphEngine`). This attribute will be immutable when the WorkGraph is finished.

## Updatable attributes
When the `WorkGraphNode` is created (in the `on_create` method), it is also stored, and its attributes become immutable. However, some of the data need updating during the execution of the WorkGraph, this includes the runtime information, e.g.,:

- state of the task
- action (e.g., pause, reset) of the task
- process of the task, because

To achieve this, we have to define the attributes that can be updated in the `_updatable_attributes` list, and save these runtime info as the attributes of the node.

## Issues
For the moment, the default logger of the `WorkGraphNode` has a `WARNING` level, however, we need `REPORT` level. I hardcoded the `logger` for the `WorkGraphNode` using the same one as in the `WorkChainNode` to fix this. We need to fix this in the future.


docs/source/development/engine.md
```diff
@@ -0,0 +1,25 @@
+# Engine
+
+In WorkGraph engine, the tasks and their inputs and outputs are passed as inputs of the WorkGraphEngine. This is different from the WorkChain, where the developers hardcode the input and output port, and the outline for every workflow.
```
agoscinski:
```suggestion
In `aiida-workgraph`, we use the `WorkGraph` object to dynamically define the inputs, outputs and outline of the workflow. It is only when the `WorkGraph` executed (run/submit) that this information is passed to the aiida engine  as `WorkGraphEngine`. In contrast to `WorkChain`, where the inputs, outputs and outline need to be all provided within the class definition, this approach allows a more interactive construction of the workflow and also allows us to retrieve information about the workflow before execution which is for example required for the visualization.
```

docs/source/development/engine.md
```diff
@@ -0,0 +1,25 @@
...
+
+When running a WorkGraph, there are the following basic steps:
+
+- export the WorkGraph as a dictionary (raw inputs), and pass it as input to the WorkGraphEngine
```
agoscinski:
```suggestion
- export the `WorkGraph` as a dictionary (raw inputs), and pass it as input to the `WorkGraphEngine`
```

docs/source/development/engine.md
```diff
@@ -0,0 +1,25 @@
...
+- a `WorkGraphNode` is created by the process
+- get the WorkGraph data from the raw inputs, and save it as attributes of the `WorkGraphNode`, this includes:
+    - WorkGraph data (takss, links, inputs, outputs etc)
```
agoscinski:
```suggestion
    - `WorkGraph` data (taks, links, inputs, outputs etc)
```

docs/source/development/engine.md
```diff
@@ -0,0 +1,25 @@
+# Engine
+
+In WorkGraph engine, the tasks and their inputs and outputs are passed as inputs of the WorkGraphEngine. This is different from the WorkChain, where the developers hardcode the input and output port, and the outline for every workflow.
+
+When running a WorkGraph, there are the following basic steps:
+
+- export the WorkGraph as a dictionary (raw inputs), and pass it as input to the WorkGraphEngine
+- a `WorkGraphNode` is created by the process
+- get the WorkGraph data from the raw inputs, and save it as attributes of the `WorkGraphNode`, this includes:
```
agoscinski:
```suggestion
- get the `WorkGraph` data from the raw inputs, and save it as attributes of the `WorkGraphNode`, this includes:
```

docs/source/development/engine.md
```diff
@@ -0,0 +1,25 @@
...
+    - WorkGraph data (takss, links, inputs, outputs etc)
+    - task states
+    - task processes
+    - task actions
+- run the WorkGraph, and update the task states, processes, and actions
```
agoscinski:
```suggestion
- run the `WorkGraph`, and update the task states, processes, and actions
```

docs/source/development/engine.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+## Updatable attributes
+When the `WorkGraphNode` is created (in the `on_create` method), it is also stored, and its attributes become immutable. However, some of the data need updating during the execution of the WorkGraph, this includes:
```
agoscinski:
```suggestion
When the `WorkGraphNode` is created (in the `on_create` method), it is also stored, and its attributes become immutable. However, some of the data need updating during the execution of the `WorkGraph`, this includes:
```

docs/source/development/engine.md
```diff
@@ -0,0 +1,25 @@
...
+
+- state of the task
+- action (e.g., pause, reset) of the task
+- process of the task, because
```
agoscinski: sentence is not complete

docs/source/development/engine.md
```diff
@@ -0,0 +1,25 @@
...
+    - task processes
+    - task actions
+- run the WorkGraph, and update the task states, processes, and actions
+
+
+## Updatable attributes
+When the `WorkGraphNode` is created (in the `on_create` method), it is also stored, and its attributes become immutable. However, some of the data need updating during the execution of the WorkGraph, this includes:
+
+- state of the task
+- action (e.g., pause, reset) of the task
```
agoscinski:
```suggestion
- action (kill, pause, play, skip) of the task
```

docs/source/development/engine.md
```diff
@@ -0,0 +1,25 @@
...
+    - task states
+    - task processes
+    - task actions
+- run the WorkGraph, and update the task states, processes, and actions
+
+
+## Updatable attributes
+When the `WorkGraphNode` is created (in the `on_create` method), it is also stored, and its attributes become immutable. However, some of the data need updating during the execution of the WorkGraph, this includes:
+
+- state of the task
```
agoscinski:
```suggestion
- state of the task (`RUNNING`, `CREATED`, `PLANNED`, `READY`)
```

docs/source/development/engine.md
```diff
@@ -0,0 +1,25 @@
...
+- action (e.g., pause, reset) of the task
+- process of the task, because
+
+
+To achieve this, we have to define the attributes that can be updated in the `_updatable_attributes` list.
```
agoscinski:
```suggestion
To achieve this, we have to define the attributes that can be updated in the `_updatable_attributes` list managed by `plumpy`.
```

src/aiida_workgraph/engine/task_manager.py
```diff
@@ -127,31 +127,37 @@ def task_set_context(self, name: str) -> None:
             result = self.ctx._tasks[name]["results"][result_name]
             update_nested_dict(self.ctx, key, result)
 
-    def get_task_state_info(self, name: str, key: str) -> str:
-        """Get task state info from ctx."""
+    def get_task_runtime_info(self, name: str, key: str) -> str:
+        """Get task state info from attributes."""
```
agoscinski, line 131: I think `runtime` would be a good naming, if aiida would no have the problem of using the terms `run` and `submit` that in this case would both refer to `runtime`. What do you think about `execution_time` to avoid this confusion?

src/aiida_workgraph/orm/workgraph.py
```diff
@@ -11,28 +11,167 @@
...
+    @property
+    def task_processes(self) -> Optional[str]:
+        """
+        Return the task process info
+
+        :returns: string representation of the task process info
+        """
+        return self.base.attributes.get(self.TASK_PROCESSES_KEY, {})
+
+    def set_task_processes(self, task_processes: str) -> None:
```
agoscinski: We already discussed, would add `property.setters`

tests/test_workgraph.py
```diff
@@ -67,11 +65,12 @@ def test_save_load(wg_calcfunction, decorated_add):
...
     # assert wg2.tasks.add1.executor == decorated_add
     # remove the extra, will raise an error
-    wg.process.base.extras.delete(WORKGRAPH_EXTRA_KEY)
-    with pytest.raises(
-        ValueError, match=f"WorkGraph data not found for process {wg.process.pk}."
-    ):
-        WorkGraph.load(wg.process.pk)
+
+
+def test_load(create_process_node):
```
GeigerJ2:
```suggestion
def test_load_failure(create_process_node):
```

src/aiida_workgraph/orm/workgraph.py
```diff
@@ -11,28 +11,167 @@
...
+
+        :returns: string representation of the workgraph data
+        """
+        return self.base.attributes.get(self.WORKGRAPH_DATA_KEY, None)
+
+    def set_workgraph_data(self, workgraph_data: dict) -> None:
+        """
+        Set the workgraph data
+
+        :param data: string representation of the workgraph data
```
GeigerJ2:
```suggestion
        :param workgraph_data: dict representation of the workgraph data
```

src/aiida_workgraph/orm/workgraph.py
```diff
@@ -11,28 +11,167 @@
...
+        Set the task action info
+
+        :returns: string representation of the task action info
+        """
+        task_actions = self.task_actions
+        task_actions[task_name] = task_action
+        return self.set_task_actions(task_actions)
+
+    @property
+    def workgraph_data(self) -> Optional[str]:
```
GeigerJ2: Doesn't that return the WG `dict` representation, or am I missing something?

src/aiida_workgraph/utils/__init__.py
```diff
@@ -332,32 +329,12 @@ def get_processes_latest(
...
-        results = qb.all()
-        # change results to dict
-        results = dict(zip(projections, results[0]))
-        # print("results: ", results)
-        for name in node_names:
-            state = results[f"extras._task_state_{name}"]
-            task_process = deserialize_safe(results[f"extras._task_process_{name}"])
+        task_states = node.task_states
+        task_processes = node.task_processes
+        task_names = [task_name] if task_name else task_states.keys()
```
GeigerJ2, line 334: Here, `task_states` seems to be a dictionary?

src/aiida_workgraph/orm/workgraph.py
```diff
@@ -11,28 +11,167 @@
...
+        return super()._updatable_attributes + (
+            cls.WORKGRAPH_DATA_KEY,
+            cls.WORKGRAPH_DATA_SHORT_KEY,
+            cls.TASK_STATES_KEY,
+            cls.TASK_PROCESSES_KEY,
+            cls.TASK_ACTIONS_KEY,
+        )
+
+    @property
+    def task_states(self) -> Optional[str]:
```
GeigerJ2: Also, here, type hint indicates return type `str`, but in `src/aiida_workgraph/utils/__init__.py` it is accessed like a dictionary?

src/aiida_workgraph/engine/workgraph.py
```diff
@@ -139,10 +146,18 @@ def save_instance_state(
...
         # Load the context
         self._context = saved_state[self._CONTEXT]
-        self.set_logger(self.node.logger)
+        # TODO: avoid hardcoding the logger
+        self.node._logger = logging.getLogger(
+            "aiida.orm.nodes.process.workflow.workchain.WorkChainNode"
+        )
+        # First time the property is called after the node is stored, create the logger adapter
+        self.node._logger_adapter = create_logger_adapter(self.node._logger, self.node)
+        self.set_logger(self.node._logger_adapter)
```
agoscinski, line 160: The assignment to the internal member variables looks hacky. Following aiida it looks like you just need to define the class variable `_logger` in `WorkGraphNode`. See inheritance
https://github.com/aiidateam/aiida-core/blob/main/src/aiida/orm/nodes/process/process.py#L260-L261
https://github.com/aiidateam/aiida-core/blob/f4c55f5f78cd7fde9a5b4a4e48cad2159fd666b2/src/aiida/orm/nodes/node.py#L140
https://github.com/aiidateam/aiida-core/blob/f4c55f5f78cd7fde9a5b4a4e48cad2159fd666b2/src/aiida/orm/utils/node.py#L157

So WorkGraphNode
```python
        _logger = logging.getLogger(
            "aiida.orm.nodes.process.workflow.workchain.WorkChainNode"
        )
```

and here
```suggestion
self.set_logger(self.node.logger)
```
The logger adapter is automatically created in `ProcessNode.get_logger` if not existent (`WorkGraphNode` inherits from `ProcessNode`)

src/aiida_workgraph/engine/workgraph.py
```diff
@@ -252,22 +267,22 @@ def on_wait(self, awaitables: t.Sequence[t.Awaitable]):
...
         """Called when a Process is created."""
         from aiida_workgraph.utils.analysis import WorkGraphSaver
 
         super().on_create()
-        wgdata = self.inputs.wg._dict
-        restart_process = (
-            orm.load_node(wgdata["restart_process"].value)
-            if wgdata.get("restart_process")
-            else None
+        self.node.label = self._raw_inputs["workgraph_data"]["name"]
```
agoscinski: Shouldn't we use here the `WorkGraphSpec.WORKGRAPH_DATA_KEY` instead of `workgraph_data`? Because later you use this to pass it  WorkGraphEngine. So if the spec key changes this will fail.

src/aiida_workgraph/orm/workgraph.py
```diff
@@ -11,28 +11,167 @@
...
+        :param data: string representation of the workgraph data
+        """
+        return self.base.attributes.set(self.WORKGRAPH_DATA_KEY, workgraph_data)
 
     @property
-    def workgraph_state_info(self) -> Optional[str]:
+    def workgraph_data_short(self) -> Optional[str]:
         """
-        Return the workgraph state info
+        Return the workgraph data
```
agoscinski: Can you give more explanation what `data_short` is in comparison to just `data`?

src/aiida_workgraph/utils/__init__.py
```diff
@@ -332,32 +329,12 @@ def get_processes_latest(
...
+        for name in task_names:
+            state = task_states[name]
+            task_process = deserialize_safe(task_processes.get(name, ""))
```
agoscinski, line 337: Note for me: Because we have now store the information in the node we can retrieve this information directly from the node instead of querying the database

src/aiida_workgraph/utils/analysis.py
```diff
@@ -61,7 +64,7 @@ def clean_hanging_links(self) -> None:
             ):
                 self.wgdata["links"].remove(link)
 
-    def save(self) -> None:
+    def analyze(self) -> None:
```
agoscinski, line 67: Can you adapt the docstring? It is still for save

tests/test_awaitable_task.py
```diff
@@ -103,7 +103,7 @@ def test_task_monitor(decorated_add):
     #
     wg1 = WorkGraph(name="wg1")
     wg1.add_task(decorated_add, "add1", x=1, y=2, t=5)
-    wg1.submit(wait=True)
+    wg1.submit(wait=True, timeout=60)
```
agoscinski, line 106: Do we raise an error if we surpass the timeout? I am just worried that the test passes due to the timeout even though it should not. Below only the creation time is checked.

agoscinski (changes requested):

src/aiida_workgraph/utils/__init__.py
GeigerJ2, line 334: `task_states` dict here, but property annotated as return `str`

src/aiida_workgraph/utils/analysis.py
```diff
@@ -80,6 +83,10 @@ def save(self) -> None:
             )
             print("modified_tasks: {}".format(modified_tasks))
```
GeigerJ2: For the future: Should we use loggers rather than print throughout?

src/aiida_workgraph/utils/analysis.py
```diff
@@ -201,9 +219,8 @@ def insert_workgraph_to_db(self) -> None:
         # pprint(self.wgdata)
         # self.wgdata["created"] = datetime.datetime.utcnow()
         # self.wgdata["lastUpdate"] = datetime.datetime.utcnow()
-        short_wgdata = workgraph_to_short_json(self.wgdata)
-        self.process.base.extras.set(WORKGRAPH_SHORT_EXTRA_KEY, short_wgdata)
-        self.save_task_states()
+        self.short_wgdata = workgraph_to_short_json(self.wgdata)
+        # self.save_task_states()
```
GeigerJ2: This is commented out? We don't set the `short_wgdata` to the `updatable_attributes` (similar to `.base.extras` before)?

src/aiida_workgraph/engine/task_manager.py
superstar54: We need an adjective here, and `runtime` can be an adjective, `execution_time` is more like a noun. I prefer `runtime`.

src/aiida_workgraph/utils/analysis.py
```diff
@@ -223,12 +239,12 @@ def save_task_states(self) -> Dict:
         task_processes = {}
         task_actions = {}
         for name, task in self.wgdata["tasks"].items():
```
GeigerJ2, line 235: Just a thought: Can we define `wgdata` as a dataclass, to have access via dot notation and to have a definition (schema) _somewhere_ what to expect that it holds? And only convert to a dictionary where needed (maybe this is everywhere, where `wgdata` is used :D). Just saying this bc I'm never sure what is inside `wgdata`, `tdata`, etc., and I always have to add breakpoints and/or print the (massive) dictionary contents to see what is contained where.

GeigerJ2 (commented):
Not familiar enough with the whole code base to give a founded review, I think. Changes look good to me. I saw @agoscinski requested changes, so I'll only comment here for now.

superstar54:
Thanks for the review. @agoscinski @GeigerJ2  . I will implement the suggestions soon.

src/aiida_workgraph/engine/workgraph.py
superstar54: Thanks for investigating this. I think I tried this solution, but it does not work. There are still something I missed. For the moment, I hardcoded the `logger`, and will open an issue after this PR, so we can fix it in the future.

src/aiida_workgraph/orm/workgraph.py
superstar54: The workgraph data is the data that only has minimal information to be displayed in the GUI. I update the docstring.

tests/test_awaitable_task.py
superstar54: I just checked, and it will raise an error when it surpasses the timeout.

src/aiida_workgraph/orm/workgraph.py
superstar54: Yes, it should be `dict`

src/aiida_workgraph/orm/workgraph.py
superstar54: Yes, it should be a `dict`

src/aiida_workgraph/utils/analysis.py
superstar54: Yes. I delete this line.

src/aiida_workgraph/utils/analysis.py
superstar54: this is moved to `insert_workgraph_to_db`. I deleted the lines.

src/aiida_workgraph/utils/analysis.py
superstar54: For most of the cases, yes.
In the WorkGraphEngine, we need to check if the customized dataclass can be serialized using aiida's default serializer, which is needed to store the checkpoint.

src/aiida_workgraph/orm/workgraph.py
superstar54: Yes, I added.

src/aiida_workgraph/engine/workgraph.py
superstar54: yes.

superstar54:
Hi @agoscinski and @GeigerJ2 , I have implemented all the review suggestions. Please take a look.

agoscinski (approved):
Thanks!

src/aiida_workgraph/engine/task_manager.py
agoscinski: Just to be clear, in `runtime info` the `runtime` is not an adjective, it is a noun adjunct (e.g. `coffee shop`) and the same way in `execution time info` the `execution time` is a noun adjunct (e.g. `coffee shop owner`). But in any case I think what you say is valid that `runtime info` more concise.

GeigerJ2 (approved):
LGTM

src/aiida_workgraph/engine/task_manager.py
agoscinski: You also use the term execution in `execute_...` in TaskManager.

PR #412, opened 2025-02-26
Save pickled data as an `updatable` attribute

### LINKED ISSUE #404: separate pickle data and source code
delete the pickled data after the workgraph is finished.

superstar54:
On top of #410 

WorkGraph allows the user to create a task from the local callables (function, WorkChain). Internally, WorkGraph pickles the local callables as binary. Previously, the binary data is stored as an AiiDA data node (PickledFunciton), which is then used as the WorkGraph input. However, the binary data could be very large, and we should provide an option to choose whether to store it or not. However, since we use it as the input, we can not delete it anymore. This is bad.

In this PR, we are not storing it as an additional AiiDA data node, but as the attributes of the WorkGraph process node. Especially, the attribute is an `updatable` attribute, which we can delete when the WorkGraph finishes (but before sealed).

## NodeExecutor
Additionally, refactor the `NodeExecutor` to keep all necessary information of the executor to be pickled, including the pickled binary data and source code of the local callable. This includes:

- task_executors
- task_error_handlers
- workgraph_error_handlers

`Error handler` is the function that runs when a node fails. It is used in the restart workflow and can recover from a failure. 

## Further improvement
- Provide the option to delete the pickled data when creating the task.

- Large binary data as an attribute will become very slow, because the attribute is stored in the database. We may switch to `repository` for performance. However, we need to update the `aiida-core` to support `updatable_objects` for a un-sealed node. Opened a PR in `aiida-core`: https://github.com/aiidateam/aiida-core/pull/6773


superstar54:
Hi @agoscinski and @GeigerJ2 , This PR is based on #410 . We need to merge #410  first and then rebase this PR. Once that is done, you can proceed with the review.

superstar54:
Hi @agoscinski  and @GeigerJ2 , ready for review :pray:

src/aiida_workgraph/engine/error_handler_manager.py
```diff
@@ -15,7 +15,8 @@ def run_error_handlers(self, task_name: str) -> None:
         if not node or not node.exit_status:
             return
         # error_handlers from the task
-        for _, data in self.ctx._tasks[task_name]["error_handlers"].items():
+        error_handlers = self.process.node.task_error_handlers.get(task_name, {})
+        for _, data in error_handlers.items():
```
GeigerJ2:
```suggestion
        for data in error_handlers.values():
```

GeigerJ2: Maybe same below, to avoid using `_` as dummy variable?

src/aiida_workgraph/engine/task_manager.py
```diff
@@ -250,7 +248,8 @@ def run_tasks(self, names: List[str], continue_workgraph: bool = True) -> None:
             self.process.report(
                 f"Run task: {name}, type: {task['metadata']['node_type']}"
             )
-            executor, _ = get_executor(task["executor"])
+            executor = self.process.node.task_executors[name]
+            executor = NodeExecutor(**executor).executor if executor else None
```
GeigerJ2:
```suggestion
            executor_dict = self.process.node.task_executors[name]
            executor = NodeExecutor(**executor_dict).executor if executor_dict else None
```
Maybe? Otherwise the second line gives me a brain aneurysm with all the `executor` :D

src/aiida_workgraph/task.py
```diff
@@ -92,9 +92,8 @@ def set_from_builder(self, builder: Any) -> None:
 
     def set_from_protocol(self, *args: Any, **kwargs: Any) -> None:
         """Set the task inputs from protocol data."""
-        from aiida_workgraph.utils import get_executor
 
-        executor = get_executor(self.get_executor())[0]
+        executor = NodeExecutor(**self.get_executor()).executor
```
GeigerJ2, line 96: One general idea: to avoid the many `NodeExecutor(**executor).executor` calls, one could add a `from_dict` `classmethod` as alternative constructor to `NodeExecutor`? Then this would simplify to:
```python
executor = NodeExecutor.from_dict(**self.get_executor())  # 
```
which might be more readable?

Also, one thing I noted when checking the implementation in `node-graph` is that `self.get_executor()` is already supposed to return a `NodeExecutor`, rather than a dict?

src/aiida_workgraph/utils/analysis.py
```diff
@@ -200,12 +200,17 @@ def insert_workgraph_to_db(self) -> None:
...
+        self.process.task_actions = self.task_actions
+        self.process.task_executors = self.task_executors
+        self.process.task_error_handlers = self.task_error_handlers
         self.process.workgraph_data = self.wgdata
         self.process.workgraph_data_short = self.short_wgdata
+        self.process.workgraph_error_handlers = self.workgraph_error_handlers
 
-    def serialize_workgraph_data(self) -> None:
+    def separate_workgraph_data(self) -> None:
         """Save a new workgraph in the database.
```
GeigerJ2: Maybe a bit more explanation in the docstring or a comment would be useful for the future?

tests/test_workgraph.py
```diff
@@ -58,9 +57,8 @@ def test_save_load(wg_calcfunction, decorated_add):
     wg2 = WorkGraph.load(wg.process.pk)
     assert len(wg.tasks) == len(wg2.tasks)
     # check the executor of the decorated task
-    callable = wg2.tasks.add1.get_executor()["callable"]
-    assert isinstance(callable, PickledFunction)
-    assert callable.value(1, 2) == 3
+    callable = wg2.tasks.add1.get_executor()
```
GeigerJ2, line 60: Top-level PR comment: I like the use of `NodeExecutor` here, as it makes it much more readable, and reduces the many accesses of (nested) dicts via str keys. Maybe we can do this for more entities?

GeigerJ2 (approved):
LGTM! Reads much nicer now, I guess as a side-effect of splitting things up and using `NodeExecutor`. Few minor comments.

src/aiida_workgraph/task.py
superstar54: The `NodeExecutor` is a dataclass, so we can pass the dict to it with **.  For the moment, using ** is fine.
```python
NodeExecutor(**data)
NodeExecutor.from_dict(data)
```
Could add `from_dict` in the future, if we have more use cases.

For the moment, the `self.get_executor()` will return a dict. A little bit mess here, we should double check and make it consistent in the future.

GeigerJ2: Yes, the dict unpacking and passing to the constructor makes sense. I thought it would read more natural with the alternative constructor. However, I just checked in the definition of `NodeExecutor`, and saw that `.executor` is not just an instance of `NodeExecutor` (as I thought), but it "Dynamically retrieves the actual callable (function/class/method/etc.) based on the NodeExecutor's mode.". So one would go from the current:

```python
executor = NodeExecutor(**self.get_executor()).executor
```

to

```python
executor = NodeExecutor.from_dict(**self.get_executor()).executor
```

so one doesn't gain in terms of readability, anyway.

Issue #408, opened 2025-02-19
Use `wg.context.x` instead of `{{x}}}`


superstar54:
## Use data from context
Old:
```python
wg.context = {"x": Int(2), "data.y": Int(3)}
wg.add_task(add, name="add1", x = `{{x}}` }
```
Proposal:

```python
wg.context = {"x": Int(2), "data.y": Int(3)}
wg.add_task(add, name="add1", x = wg.context.x }
```

## Update data in context 
Old:
```python
wg.context = {"x": Int(2), "data.y": Int(3)}
add1 = wg.add_task(add, name="add1", x = wg.context.x }
add1.set_context({"sum": "result"})
```

Proposal:

```python
wg.context = {"x": Int(2), "data.y": Int(3)}
add1 = wg.add_task(add, name="add1", x = wg.context.x }
add1.set_context({wg.context.x: "result"})
```

Or even better if we can assign the socket to the context variable directly:

```python
wg.context = {"x": Int(2), "data.y": Int(3)}
add1 = wg.add_task(add, name="add1", x = wg.context.x }
wg.context.sum = add1.output.result
add2 = wg.add_task(add, name="add2", x = wg.context.x }
wg.context.sum = add2.output.result
```

**This should not override the value of the sum**, but only register that the value will be updated when the `add1` task (or `add2` task) is finished. Thus, when printing the `sum`, it should give the following:
```
[add1.output.result, add2.output.result]
```

superstar54:
Implemented in #433

PR #415, opened 2025-02-27
WIP implementing tracked context variables

agoscinski:
This PR is working on issue issue #408.

For graphical visualization in the GUI we need some class that tracks the connection that the context variable is used for. The context itself is completely agnostic of its access by design. Also the context is an object of plumpy and is completely unaware of the GUI, therefore in workgraph we use the `node_graph.socket.NodeSocket` for this purpose. These are heavily coupled with `node_graph.link.NodeLink` that tracks it connection to other nodes.

For this implementation we need to create a new type of Sockets with extended capabilities:

1. We need to track usage and create new sockets internally while the user is not necessary aware of it.
```python

def increment(x): return x+1

wg.context.x = aiida.orm.Int(5) # create socket x0 with initial value
t = wg.add_task(increment, x = wg.context.x) # link socket x0 to t.inputs.x
wg.context.x = t.outputs.result # create socket x1 and link to results
t2 = wg.add_task(increment, x = wg.context.x) # link socket x1 to t2.inputs.x
```

2. We need to allow bidirectional usage as input and output
```python
wg.add_link(wg.context.x, t.inputs.x)
…
wg.add_link(t.outputs.x, wg.context.x)
```

Since a socket by itself is not containing any logic how it is populated, we need also a new `Task` class specifically for managing the `Socket`. We need to at least implement a wrapper around the sockets to implement custom `__getattr__`, and `__setattr__` to create Sockets on access. At first glance, a specialized `NodeSocketNamespace` looks like the the right way to implemnt this, but on the other hand we want to implement a very specfic access pattern that that easily can interfere with the other patterns of `NodeSocketNamespace`. So it seems better to implement a separated `AccessInterface` in `WorkGraph` or the specialized `Task` that exposes this `NodeSocketNamespace` to the user.

This requires to touch part multiple parts of the code. As we need to implement has this new `Task` and `Socket` is handled during execution (population of socket, and passing of this new socket namespace). The population of the socket is outline in this WIP but the new socket namespace is not passed to the execution time.

Furthermore, how this is implemented at the GUI is still an open question. We have a design ideas, see images in the PR on GitHub.

### Design ideas for GUI

The visualization of the aiida.orm.Int nodes is just to clarify it and is not part of this PR

#### Example 1
```python
def increment(x): return x+1

wg.context.x = aiida.orm.Int(2)
wg.add_task(increment, x = wg.context.x)
```

image

#### Example 2

```python
def increment(x): return x+1

wg.context.x = aiida.orm.Int(5)
t = wg.add_task(increment, x = wg.context.x)
wg.context.x = t.outputs.result

wg.add_task(increment, x = wg.context.x)
```

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -128,6 +128,8 @@ workgraph = "aiida_workgraph.cli.cmd_workgraph:workgraph"
 [project.entry-points."aiida_workgraph.socket"]
 "workgraph.any" = "aiida_workgraph.sockets.builtins:SocketAny"
 "workgraph.namespace" = "aiida_workgraph.socket:TaskSocketNamespace"
+"workgraph.variables" = "aiida_workgraph.socket:ContextSocketNameSpace"
+"workgraph.variable" = "aiida_workgraph.sockets.builtins:SocketVariable"
 "workgraph.int" = "aiida_workgraph.sockets.builtins:SocketInt"
 "workgraph.float" = "aiida_workgraph.sockets.builtins:SocketFloat"
 "workgraph.string" = "aiida_workgraph.sockets.builtins:SocketString"
--- a/src/aiida_workgraph/engine/task_manager.py
+++ b/src/aiida_workgraph/engine/task_manager.py
@@ -187,6 +187,7 @@ class TaskManager:
         return is_finished, result
 
     def continue_workgraph(self) -> None:
+        breakpoint()
         self.process.report("Continue workgraph.")
         task_to_run = []
         for name, task in self.ctx._tasks.items():
@@ -245,6 +246,7 @@ class TaskManager:
                 f"Run task: {name}, type: {task['metadata']['node_type']}"
             )
             executor, _ = get_executor(task["executor"])
+            breakpoint()
             args, kwargs, var_args, var_kwargs, args_dict = self.get_inputs(name)
             for i, key in enumerate(self.ctx._tasks[name]["args"]):
                 kwargs[key] = args[i]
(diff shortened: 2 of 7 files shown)


agoscinski:
Considering now that we want to edit the node graph widget in some future, the above designs do not allow to go back from the graph widget representation to the actual workflow as you completely lost the order of execution. One solution would be to also store the order of the links but that seems hard to read. I think it is therefore better to also split the variable usage into separate tasks like in the example below. 

```python
def increment(x): return x+1
def compare(x): return x<10


wg.context.x = aiida.orm.Int(5)
t = wg.add_task(compare, x = wg.context.x)
wg.context.x = t.outputs.result

wg.add_task(increment, x = wg.context.x)
```

image

This would imply for the core representation in `aiida-workgraph` that it should create a new `ContextTask` (the new custom task discussed in the first post) for each variable assignment.

superstar54:
> the above designs do not allow to go back from the graph widget representation to the actual workflow as you completely lost the order of execution. 

Hi @agoscinski , this will not be a problem, because of possible cyclic dependence, input link from a context variable will not account for the execution order. In case the order is not defined because the use of context variable, the user needs to use the `_wait` link to set the order explicitly.

agoscinski:
Just further elaborating @superstar54 for me. Creating a new Task for each new assignment is problematic as it is not clear when to reuse the input and when to create a new assignment. In the above example I did not include an assignment at the end which would be ambiguous
```python
def increment(x): return x+1
def compare(x): return x<10

wg.context.x = aiida.orm.Int(5) # new assignment creates a socket x0
t = wg.add_task(compare, x = wg.context.x)
wg.context.x = t.outputs.result # new assignment creates a socket x1

increment = wg.add_task(increment, x = wg.context.x)
wg.add_link(increment.outputs.result, wg.context.x) # assignment creates a socket x2 but we want to link to x0
```
Therefore for now we need can only use one task for the representation and continue to use the aiida context in the background. The order for now will be determined by the waits the user has to provide. In the future we can think about this case again, maybe for these special logic (if-then-else, while and for loops) we can handle the `ContextTask`/`ContextSocket` (whatever we implement here) differently.

superstar54:
Closed by #433

PR #416, opened 2025-02-28
Improved makefile for docs build

GeigerJ2:
Some improvements to building the docs locally.

Added additional `make` (`sphinx-build`) targets `incremental` and `livehtml` (names should be self-explanatory), explicit `clean` operation, rather than catch-all pattern, as well as `file` target. Using `-j auto` to use all available CPU cores the default.

Changes done mainly by Claude.ai, so I hope this makes sense 😅 works at least locally for me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,10 +133,11 @@ dmypy.json
 # Pyre type checker
 .pyre/
 
-#
+# Misc
 *.pdf
 tests/work
 /tests/**/*.png
 /tests/**/*txt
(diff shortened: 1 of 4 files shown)


superstar54 (approved):
LGTM!

GeigerJ2:
Hi @superstar54, I saw that the CI failed because of the `test_reset_message` test. IIRC you mentioned in the past that this one is unreliable? Or is that actually blocking the merge?

PR #426, opened 2025-03-07
Add Map tasks

superstar54:
Similar to the `if_` and `while_` instructions, this PR adds a `Map` zone task and a `map_` instruction.

The `map_` instruction in WorkGraph is conceptually similar to Python's built-in map function. The `map_` instruction spawns the tasks in the Map zone for each item in a dataset.

### Why use `map_`?

- **Creates dynamic tasks**: The number of tasks is determined at runtime based on input size.
- **Parallel execution**: Tasks inside a `map_` block (Map zone) run simultaneously and independently for each input item.

### Example

```python
from aiida_workgraph import task, WorkGraph, map_

load_profile()

@task.calcfunction(outputs=["result"])
def generate_data(n: int) -> dict:
    """Generate a dictionary of integers."""
    return {"result": {f"key_{i}": orm.Int(i) for i in range(n.value)}}

@task.calcfunction()
def add(x, y):
    return x + y

@task.calcfunction()
def calc_sum(**datas) -> dict:
    """Compute the sum of all provided values."""
    return  sum(datas.values())

n = 3
wg = WorkGraph("parallel_addition")
wg.add_task(generate_data, name="generate_data", n=n)
# for each item in the `result` dict, run the `add` task
map_(wg.tasks.generate_data.outputs.result)(
    # the input x of the `add` task is replaced by the item
    wg.add_task(add, name="add", x=map_.default_placeholder, y=1),
)
wg.add_task(calc_sum, name="calc_sum", datas=wg.tasks.add.outputs.result)
wg.run()
print("result: ", wg.tasks.calc_sum.outputs.result)
```
The GUI representation of the WorkGraph:
Screenshot from 2025-03-08 18-04-23

This will generate the following node graph provenance:
image


superstar54:
Hi @rikigigi, the `map` feature has now been added! 🎉 Ping @GeigerJ2 and @agoscinski to keep you both in the loop.

A simple example is included in the PR description—if you have any feedback on the syntax, feel free to share your thoughts.

@GeigerJ2, the EOS workflow can use this `map_` instruction; thus, there is no need to use the `graph_builder` anymore.

Since this is still an intermediate step in refactoring the engine, I’ll proceed with merging the PR without a review.

PR #392, opened 2024-12-16
WIP: Create `Task` from `builder`

GeigerJ2:
- [x] Tests
- [x] Docs

PR #432, opened 2025-03-12
Add feature to create Task from AiiDA ProcessBuilder

GeigerJ2:
Superseeds #392, using the new `AiiDAComponentTaskFactory`.


superstar54:
The tests show there are multiple `add@localhost` code, maybe in the fixture
```
@pytest.fixture
def add_code(fixture_localhost):
```
We need to load the code first, if it does not exist, we create one. What do you think?

GeigerJ2:
Yes, I just realized the failure and applied this fix 👍🏽

superstar54 (approved):
Nice!

PR #433, opened 2025-03-12
Implements context namespace with accessor for syntactic sugar

agoscinski:
Implementation a namespace for context variables `ContextSocketNamespace`. Does not contain much logic for the moment. But in future work more logic is needed here to manage its handling in the engine.

The `ContextSocketNamespaceAccessor` is an access wrapper around the `ContextSocketNamespace` for syntactically nicer usage hiding complexity of the socket namespace from the user.

In principle for the current approach the creation sockets and namespaces is not needed, since it just converted into to the curly brackets dicts when passed to the engine. At some point we however need these custom objects to represent them in the GUI. To reduce the complexity of the approach, this PR does only bring the syntax wrapper around the context namespace.

Example usage
``` python
wg = WorkGraph()
wg.context["data.x"] = 5
# same as  
wg = WorkGraph()
wg.context = {"data.x": 5}
# same as  
from aiida_workgraph.socket import ContextSocketNamespace
wg = WorkGraph()
wg.context.data = ContextSocketNamespace("data") # will raise error if names do not match
wg.context.data.x = 5
```

Context can be passed
```
load_profile()
wg = WorkGraph()
wg.context['data.x'] = None # creates internally a namespace 'data' socket 'x' and sets sockets value to "{{data.x}}"
print(wg.context.data.x) # '{{data.x}}'
@task.calcfunction
def incr(x): return x+1
incr_t1= wg.add_task(incr, "incr1", x=5)
incr_t1.set_context({wg.context.data.x: "result"}) # internally the curly brackets are stripped away when passed to the engine
incr_t2 = wg.add_task(incr, "incr2", x=wg.context.data.x)
incr_t2.waiting_on.add(incr_t1)
wg.run()
print(incr_t2.outputs.result)
```


superstar54:
Hi @agoscinski  thanks for the work. Here are some initial suggestions:

- Support tab completion to inspect the variables inside the context, for the moment, it shows the methods instead of the variables. So better hide all methods, and show the data only.
Screenshot from 2025-03-14 15-34-26
- Better to move the class definition to a separate file.

Could you implement these changes, and fix the failed tests?

superstar54:
@agoscinski , after reconsidering our requirements, I realized that our approach to handling context variables differs from how Sockets are used. Specifically, we aim to support directly setting the value of a context variable, such as:

```python
wg.context.x = 5
```
But in the Socket, we actually **do not** support this. Instead, in a Socket, we explicitly require using the value attribute:

```python
task1.inputs.x.value = 5
# or task1.inputs.property.x.value = 5
```
The reason to do this is because we want to use the Socket in two ways:
- store the value
- store the link with the socket

And, the above two functionalities are also needed for the context. Therefore, I would suggest **not** to supporting direct attribute assignment like:
```python
wg.context.x = 5
```
instead, we should enforce using the value attribute, just like Sockets:
```python
wg.context.x.value = 5
```
This keeps the API consistent and simplifies development. 

## Initialization of Context Variables
Setting up the context like:
```python
wg.context = {"x": 1, "y": 2}
```
should internally perform:
```python
# Create and initialize sockets for each variable
socket_x = wg._context._new_("workgraph.any", "x")
socket_x.value = 1

socket_y = wg._context._new_("workgraph.any", "y")
socket_y.value = 2
```
## Updating context variables
### assig a value
```python
wg.context.x.value = 5
# or use set method
wg.update_context({"x": 5}) 
# Raise error if
wg.context.x = 5 # ❌ This should raise an error
```

### link to an output socket
If the assigned value is another Socket (e.g., an output from a task), a link should be created instead of simply overriding the value.

```python
# use set method
wg.update_context({"x": task1.outputs.result})
```
Internally, this should establish a link rather than setting a static value:
```
wg.add_link(task1.outputs.result, wg.context.x)
```

### Handling links
- One context socket can have many links.
- Order of adding the link does not matter; the context value updates based on task execution.
- The context variable acts as both an input and output socket:
    - When linked to an input socket, it passes values into a task.
    - When linked to an output socket, it stores values from task execution.

## Nested context variable
```python
wg.update_context({"task1.x": task1.outputs.result})
```
The `NodeSocketNamespace` can handle the nested context variable automatically.

In this way, we only need the `NodeSocketNamespace`, and treat the context variables as sockets. This keeps the API consistent among context, inputs and outputs., also simplifies development.

agoscinski:
We however also allow assignment of inputs like this
```python
t = wg.add_task(add, x=1, y=5)
print(t.inputs.x.value) # 1
```
So doing this
```python
wg.context.x = 5
```
is not that far away from the current syntax. But I can see the confusion with trying to mess around with the assignment operator so maybe we make everything a bit more transparent to the user by only allowing ~read/~write (EDIT I meant only write) of context through a function
```python
wg.update_context(x=5) # same as wg.update_context({"x":5})
wg.update_context({"task.x": 5}) # nested
#wg.context.x = 5 # this is not proposed to the user and also will fail
#wg.context = {"x": 1, "y": 2} # this is not proposed to the user and also will fail
```
EDIT: It is still a bit misleading since `foo(x=5)` is not the same as `foo({'x':5})`, but rather `foo(**{'x':5})`. If we would be consistent and not support `foo({'x':5})` but only `foo(**{'x':5})`, one still can do nested arguments by `foo(**{'data.x':5})`, but it is not maybe not so clear. I feel like nested arguments should be covered differently, @superstar54 what are use cases for nested arguments? I only know that for aggregation it is important, are there any other use cases? Maybe if we support the nested arguments use cases differently we can remove their usage in the context and simplify the syntax here.

### Initialization of Context Variables

```python
wg.context = {"x": 1, "y": 2}
```
This works currently. and it should do this internally what you described (using the socket namespace). But maybe it is better to not allow it and only allow
```
wg.update_context({"x": 1, "y": 2})
```

### Updating context variables
```python
wg.context.x.value = 5
```
Yes this works.

### Links
There will be no linking support in this PR. This requires the change of a lot of internal stuff and a ContextSocket class. The idea of this PR was to have a simpler goal than PR #415.

superstar54:
## Nested
Nest port is common in AiiDA's process, both inputs and outputs can be nested, e.g., in CalcJob and WorkChain. For nested inputs, we can not run this anymore
```python
t = wg.add_task(add, x=1, "data.x"=5)
```
One need to:
```python
t.set({"data.x": 5})
```

thus, I think it's fine that we only support this syntax for context, keep the API consistent.
```python
wg.update_context({"task.x": 5})
```



## Links
I think the link is supported automatically if using `NodeSocketNamespace`, for example,

```python
    @property
    def context(self) -> NodeSocketNamespace:
        if self._context is None:
            self._context = NodeSocketNamespace(name="context",
                                                pool=SocketPool,
                                                metadata={"dynamic": True})
            # Add a temporary node, but not in the WorkGraph
            # we should disable users adding task with name `_context`
            self._context._node = Task(name="_context", parent=self)
        return self._context
    
    @context.setter
    def context(self, value: Dict[str, Any]) -> None:
        self.context._clear()
        self.context._value = value
    
    def update_context(self, value: Dict[str, Any]) -> None:
        self.context._set_socket_value(value)
```

run this tests
```python

# define add task
@task.calcfunction
def add(x, y):
    return x + y

wg = WorkGraph()
wg.context = {"x": 1, "data.x": 5}
assert wg.context.x.value == 1
wg.update_context({"data.y": 10})
assert wg.context.data.y.value == 10
wg.add_task(add, name="add1", x=wg.context.x, y=wg.context.data.y)
assert len(wg.links)  == 2

```

agoscinski:
Agreed on the syntax
```python
wg.update_context({"x":  5}) # creates wg.context.x
wg.update_context({"data.x":  5}) # creates wg.context.data.x
```

```python
wg.add_link(add1.outputs.result, wg.context.sum)
#or
wg.update_context({"sum": add1.outputs.result})
```
For this we need to support the linking of the context. Since linking requires at the moment a node (so it can later be properly serialized in the wgdata) the above solution adds a task node to the namespace
```python
            self._context._node = Task(name="_context", parent=self)
```
but in the case of the context we pass it anyway separately to the wgdata so one can make a context namespace that does not need a node to be associated with. Maybe it is still easier to implement a ContextNode that is associated with a namespace and the node stores all the links. My guess is that this is more consistent with the remaining logic in the code.

superstar54:
Hi @agoscinski , I implemented the syntax, and all the tests passed.

Since this is still in the intermediate stage of the refactoring, I am going to merge this PR without review.  Sorry that I am doing this in a hurry. I want to put all the pieces of the new syntax together; then we can test it as a whole.

But feel free to try and give your feedback anytime!

PR #436, opened 2025-03-16
Pydantic schema for WorkGraphData

agoscinski:
I used
```bash
datamodel-codegen --input src/aiida_workgraph/schemas/aiida_workgraph.schema.json --input-file-type jsonschema --output models.py
```
to create a pydantic model from the existing schema. I would change everywhere in the code the usage of the dict to the usage of the pydantic model. When passed in WorkGraph to the WorkGraphEngine the model is dumped to dict, then it is converted to the pydantic model again in WorkGraphEngine.

### DIFF
--- a/src/aiida_workgraph/schemas/workgraph_data.py
+++ b/src/aiida_workgraph/schemas/workgraph_data.py
@@ -1,0 +1,139 @@
+from __future__ import annotations
+
+from typing import Any, Dict, List, Optional
+from uuid import UUID
+
+from pydantic import BaseModel, Field, conint, constr
+
+# TODO constr is deprecated
+
+class Metadata(BaseModel):
+    graph_type: Optional[str] = Field(
+        None, description='Type of this graph, e.g. NORMAL'
+    )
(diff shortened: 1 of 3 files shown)


superstar54:
Hi @agoscinski , thanks for the work!

Currently, I think the pedantic model will be helpful in two places:
- `from_yaml,` where we create the WorkGraph instance from a user-provided yaml file, e.g., comes from another engine or the common workflow.
- in the workgraph's web application where fastapi is used. In this case, maybe we need to directly update the `WorkGraphNode` to a pedantic model. This maybe similar to the on-going work in the aiida-core (restapi)?  The web application can be found in the `aiida-workgraph-web-ui` repo, e.g., in this file: https://github.com/aiidateam/aiida-workgraph-web-ui/blob/main/aiida_workgraph_web_ui/backend/app/workgraph.py

Note that the engine already dropped the usage of the `wgdata` and re-loaded the WorkGraph instance directly.

Please go ahead, maybe first update the `from_yaml` method, and then we can first merge the PR.

PR #437, opened 2025-03-17
Add graph manager and task wrapper

superstar54:
Bump `node-graph` to `v0.1.23`.

## Refactor decorators to return the wrapper instead of executing the function.
-  when called, adds a task to the current graph and returns the outputs.
- input argument mapping from `*args, **kwargs` to task inputs.
- handling of single vs. multiple outputs.

## Add manager for graph and zone
- create a global graph manager
- automatic task instantiation in the active graph and zone
 

## Implemented operator overloading for Socket to support:
- Arithmetic: +, -, *, /, //, %, **
- Comparison: <, >, <=, >=, ==, !=

Additionally, cleaned unused notebooks in the docs.


superstar54:
Hi @GeigerJ2 and @agoscinski , this PR is still an intermediate stage for the new syntax, which we discussed yesterday, I will merge this PR without review, as there may be other refactoring work soon.

For a quick knowledge:
- now run the decorated function will return the output socket, instead of executing the original function
- socket can be manipulated directly for basic binary operation (+, -, > etc), e.g., `socket1 > 0`, which will create a task automatically.

- the `active_graph` and `active_zone` is introduced

Here is a simple example:
```python
from aiida_workgraph.manager import active_graph

load_profile()


@task.calcfunction()
def add(x, y):
    return x + y

@task.calcfunction()
def multiply(x, y):
    return x * y


with active_graph(WorkGraph()) as wg:
    sum = add(x=2, y=3)
    product = multiply(x=sum, y=4)
    wg.run()
    print("result:", product.value)
```

Issue #260, opened 2024-08-22
Awaitable task is not listed in `verdi process list`


agoscinski:
The code
```python
import asyncio

@task.awaitable()
async def some_sleep():
    await asyncio.sleep(10)
    return orm.Int(0)
    
@task.calcfunction
def some():
    return orm.Int(0)
wg = WorkGraph()

some_sleep_task = wg.add_task(some_sleep, name='some_sleep_task')
some_task = wg.add_task(some, name='some_task')

wg.run()
```
Gives me first the output
```
2496  2s ago     WorkGraph<WorkGraph>                                        ⏵ Waiting         Waiting for child processes: some_sleep_task
2497  0s ago     some                                                        ⏹ Finished [0]
```
Then when everything is finished
```python
2496  13s ago    WorkGraph<WorkGraph>                                        ⏹ Finished [0]
2497  12s ago    some                                                        ⏹ Finished [0]
```
I am missing the `some_sleep_task`

superstar54:
Yes, only AiiDA processes are listed. The following tasks have a process:
- calcfunction
- workfunction
- calcjob
- workchain
- workgraph
- graph_builder who generate a workgraph

Other tasks will not:
- normal task created with `@task()`
- awaitable
- monitor
- while and if

superstar54:
But you can still visualize the state of the non-AiiDA process tasks in the GUI using web app.

superstar54:
The `task` decorator can handle async functions now, and the task is tracked by the AiiDA process, so the `awaitable` task is deleted. The `monitor` task also has its corresponding AiiDA process now.

So I am closing this issue.

PR #454, opened 2025-03-24
Use `pyfunction`

### LINKED ISSUE #338: Support store_provenance in normal task
There is no provenance stored for a `normal` task. If the user wants to store the provenance, they need to switch to `calcfunction` or `PythonJob`. But the downsides are
- `calcfunction` needs AiiDA nodes as input and output
- `PythonJob` requires the module to be importable in the working computer.

Is it possible to combine the upsides of the above tasks: 1) can be run locally; 2) automatically serialize inputs and outputs. Maybe we could add `store_provenance` argument for a `normal` task, e.g.,

```python
@task()
def add(x, y):
    return x + y

wg = WorkGraph()
add_task = wg.add_task(add, name="add1", x=1, y=2,
                       store_provenance=True,)
wg.run()
```

superstar54:
Previously, no provenance was stored for a `normal` task decorated using `@task()`. If the user wants to store the provenance, they need to switch to `calcfunction` or `PythonJob`. But the downsides are
- `calcfunction` needs AiiDA nodes as input and output
- `PythonJob` runs on a remote computer by uploading/retrieving files, which is overkill for a function that only needed to run locally.

Since `aiida-pythonjob v0.2.2`, a new `pyfunction` is added, which runs the function locally, as well as serializes the inputs/outputs like the PythonJob.

In this PR, we use this `pyfunction` for the normal `task`. So internally, it will transfer the function into a `pyfunction`.

```python

load_profile()

@task()
def add(x, y):
    return x + y

wg = WorkGraph()
add_task = wg.add_task(add, name="add1", x=1, y=2)
wg.run()
```

Then, one can check the processes
```
$ verdi process list -a
58722  11s ago    WorkGraph<WorkGraph>            ⏹ Finished [0]
58725  10s ago    add                             ⏹ Finished [0]

```

## ToDo
The following tasks are still not stored in the provenance graph:
- awaitable
- monitor
- zone tasks (while, if, map)

See this issue: https://github.com/aiidateam/aiida-workgraph/issues/260


superstar54:
Ping @GeigerJ2 and @agoscinski , so that you are aware of this important change.

PR #470, opened 2025-05-05
Add meta tasks for `group_inputs`, `group_outputs` and `ctx`

superstar54:
`node-graph` has an API that lets the user define the `graph-level` inputs and outputs. In this PR and https://github.com/scinode/node-graph/pull/52, I refactored the code to make `group_inputs`, `group_outputs` and `ctx` have the same interface.

- add meta tasks `group_inputs`, `group_ouputs` and `ctx` to hold the sockets for `group_inputs`, `group_outputs` and `ctx.
- only expose the used sockets to the user.
  - for the `group_inputs` task, it is its `outputs`
  - for the `group_outputs` task, it is its `inputs`
  - for the `ctx` task, in principle, we should create a special socket that can act as both `inputs` and `outputs`, but for simplicity, I use the `inputs`.

```python

load_profile()

@task()
def add(x, y):
    return x + y

wg = WorkGraph()
# define group inputs "a", which links to the input x of the add1 and add2 tasks
wg.group_inputs.a = 1
wg.add_task(add, name="add1", x=wg.group_inputs.a, y=1)
wg.add_task(add, name="add2", x=wg.group_inputs.a, y=wg.tasks.add1.outputs.result)
# define group outputs "sum" which links to the output of the add2 task
wg.group_outputs.sum = wg.tasks.add2.outputs.result
wg.run()
print(wg.group_outputs.sum.value)
```

The GUI representation of the above WorkGraph is:
image


src/aiida_workgraph/workgraph.py
```diff
@@ -65,6 +65,11 @@ def tasks(self) -> TaskCollection:
         """Add alias to `nodes` for WorkGraph"""
         return self.nodes
 
+    @property
+    def meta_tasks(self) -> dict:
+        """Meta tasks are the context and group inputs/outputs"""
```
GeigerJ2, line 70: Why is the context also a meta task?

src/aiida_workgraph/workgraph.py
```diff
@@ -315,25 +312,22 @@ def update(self) -> None:
...
+        if self.process.is_finished_ok:
+            # update the output sockets
+            for socket in self.group_outputs:
+                if socket._identifier == "workgraph.namespace":
+                    socket._value = get_nested_dict(
+                        self.process.outputs, socket._name, default=None
+                    )
+                else:
+                    socket.value = get_nested_dict(
+                        self.process.outputs, socket._name, default=None
```
GeigerJ2, line 324: What's the difference between `_value` and `value` here?

src/aiida_workgraph/workgraph.py
```diff
@@ -315,25 +312,22 @@ def update(self) -> None:
...
-            self._ctx = ContextSocketNamespace(
-                name="ctx", pool=SocketPool, graph=self, metadata={"dynamic": True}
-            )
-        return self._ctx
-
-    @ctx.setter
-    def ctx(self, value: Dict[str, Any]) -> None:
-        self.ctx._clear()
-        self.ctx._set_socket_value(value, link_limit=100000)
-
```
GeigerJ2, line 336: Were the changes to the `ctx` strictly necessary related for this PR, or a side-effect of some general refactoring?

src/aiida_workgraph/tasks/factory/pythonjob/tasks.py
```diff
@@ -44,7 +44,7 @@ def _serialize_python_data(cls, input_sockets: Dict[str, Any]) -> None:
         values to AiiDA Data nodes, if needed.
         """
         for sock in input_sockets.values():
```
GeigerJ2: sock🧦 :D

src/aiida_workgraph/tasks/factory/workgraph_task.py
```diff
@@ -39,23 +39,13 @@ def links(self):
...
-            # which means the data is grouped by the task name
-            for socket_name, value in data.items():
-                input = wgdata["tasks"][task_name]["inputs"]["sockets"][socket_name]
-                if input["identifier"] == "workgraph.namespace":
-                    input["value"] = value
-                else:
-                    input["property"]["value"] = value
+        # update the workgraph inputs by the kwargs
+        for socket, data in kwargs.items():
+            input = self.workgraph.group_inputs[socket]
```
GeigerJ2: This will shadow the builtin `input`. Consider changing to `input_`.

src/aiida_workgraph/tasks/factory/utils.py
```diff
@@ -26,7 +26,7 @@ def generate_tdata(
     )
     for input in task_inputs:
```
GeigerJ2: Also here shadowing the bulti-in `input`

tests/conftest.py
```diff
@@ -208,13 +208,14 @@ def add_multiply(x, y, z):
 def decorated_add_multiply_group(decorated_add, decorated_multiply) -> Callable:
     """Generate a decorated node for test."""
 
-    @task.graph_builder(outputs=[{"name": "result", "from": "multiply1.result"}])
+    @task.graph_builder(outputs=[{"name": "result"}])
```
GeigerJ2, line 211: Not entirely sure now, how the linking is resolved, such that the right task output is linked to the global "result"? But I guess this is done in the `generate_group_outputs` and `generate_group_inputs` in the node-graph PR.

tests/test_ctx.py
```diff
@@ -65,10 +65,9 @@ def add(x, y):
     wg = WorkGraph()
     wg.add_task(add, name="add", x=1, y=2)
     wg.update_ctx({"data.sum": wg.tasks.add.outputs.results.sum})
-    wg.group_outputs = [
-        {"name": "sum", "from": "ctx.data.sum"},
-        {"name": "add.sum", "from": "ctx.data.sum"},
-    ]
+    wg.group_outputs.sum = wg.ctx.data.sum
+    wg.group_outputs.add = {}
+    wg.group_outputs.add.sum = wg.ctx.data.sum
```
GeigerJ2: Not possible to do in one line?

tests/test_shell.py
```diff
@@ -116,6 +116,7 @@ def parser(dirpath):
                 {"identifier": "workgraph.any", "name": "result"}
             ],  # add a "result" output socket from the parser
         )
+        wg.group_outputs.result = wg.tasks.job2.outputs.result
```
GeigerJ2: I still think it would be nice to have syntactic sugar like `wg.outputs`. Or maybe rename to `global_inputs` and `global_outputs`? I think this is more intuitive than `group_inputs` and `group_outputs`.

tests/test_workgraph.py
```diff
@@ -178,10 +178,7 @@ def test_extend_workgraph(decorated_add_multiply_group):
 def test_workgraph_group_outputs(decorated_add):
     wg = WorkGraph("test_workgraph_group_outputs")
     wg.add_task(decorated_add, "add1", x=2, y=3)
-    wg.group_outputs = [
-        {"name": "sum", "from": "add1.result"},
-        # {"name": "add1", "from": "add1"},
-    ]
+    wg.group_outputs.sum = wg.tasks.add1.outputs.result
```
GeigerJ2: Overall, I like that such a direct assignment is now possible, rather than having to rely on lists of dictionaries 👍🏽

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -111,25 +111,26 @@ def add_multiply(x=None, y=None, z=None):
...
-# We can expose the outputs of the tasks as the outputs of the WorkGraph:
+# We can expose the outputs of the WorkGraph as the outputs of the task.
 #
 # .. code-block:: python
 #
-#     @task.graph_builder(outputs = [{"name": "multiply", "from": "multiply.result"}])
+#     @task.graph_builder(outputs = [{"name": "multiply"}])
 #
-# This will expose the `result` output of the `multiply` task as the `multiply` output of the WorkGraph.
+# This will expose the `result` output of the workgraph as the `multiply` output of the task.
```
GeigerJ2: Isn't it the other way around, i.e., exposing the output of a task as the output of a WorkGraph?

docs/gallery/tutorial/autogen/zero_to_hero.py
```diff
@@ -367,15 +368,16 @@ def add_multiply_if_generator(x, y):
...
 
-wg.submit(
+wg.run(
     inputs={
         "add1": {"x": 1, "y": 2},
         "add_multiply_if1": {"y": 2},
         "add2": {"y": 2},
     },
-    wait=True,
+    # wait=True,
```
GeigerJ2:
```suggestion
```

GeigerJ2 (commented):
OK, went once through everything. Thanks for fixing it so quickly, @superstar54! My main pet-peeve is the name. I'd propose only `outputs`/`inputs` of the WG, as that is what users will expect, or `global_outputs`/`global_inputs`, if there is a problem with using `outputs`/`inputs`. Will try this out for the project where I need it, and report back.

src/aiida_workgraph/workgraph.py
superstar54: `ctx`, `inputs`, and `outputs` are sockets. In Workgraph, a `socket` should be attached to a `task`.

src/aiida_workgraph/workgraph.py
superstar54: Because we support auto-completion using `tab` to inspect the sockets inside a `namespace socket`, so its `value` attribute is renamed to `_value`. For a normal socket, we still use `value.

src/aiida_workgraph/workgraph.py
superstar54: Yes, this is a general refactoring to make `ctx`, `inputs` and `outputs` have the same API (meta tasks).

src/aiida_workgraph/tasks/factory/pythonjob/tasks.py
superstar54: I changed it to full name `socket`. :laughing:

src/aiida_workgraph/tasks/factory/workgraph_task.py
superstar54: you are right! I changed it to `input_socket`

tests/conftest.py
superstar54: This is done inside the function
```python
wg.outputs.result = wg.tasks.multiply1.outputs.result
```

tests/test_ctx.py
superstar54: Not supported yet.
The first line defines a `namespace` output `add`.
The second line defines the output `sum` inside the namespace.

tests/test_shell.py
superstar54: I agree — using wg.inputs and wg.outputs is more user-friendly. I've updated the code accordingly as suggested.

Just a note: the original names group_inputs and group_outputs were chosen to avoid potential conflicts in case a user defined a task named inputs or outputs. For this reason, in the meta_tasks that hold the inputs and outputs sockets, I use _inputs and _outputs to avoid name collisions.

docs/gallery/howto/autogen/graph_builder.py
superstar54: No, this is a graph builder task, which we need to expose the outputs of the workgraph as the outputs of the graph builder task. I updated to 
```
# This will expose the `multiply` output of the workgraph as the `multiply` output of the task.
```

superstar54:
Hi @GeigerJ2, thanks for the review. I've updated the code based on your suggestions.

I've renamed the properties to `wg.inputs` and `wg.outputs`. The only remaining question is how to display the `meta_tasks` in the GUI. For now, I'm using `_inputs` and `_outputs` as the task names to avoid conflicts if a user creates tasks named `inputs` or `outputs`.

In the future, we may consider switching to `graph_inputs` and `graph_outputs`, and reserve those names to avoid similar name collisions.

tests/test_shell.py
GeigerJ2: Sounds good!

superstar54:
Hi @GeigerJ2 , I am merging this PR. We can improve it in the future PR if you have any suggestions.

PR #473, opened 2025-06-06
Debug RTD 2.7.0rc1 issue

agoscinski:
TODO:
- [ ] to separate PR for adding post release to aiida-workgraph. Reasoning:
>  It is better to mark the repo version as <VERSION>post after a release because then you do not accidentally download the version from pypi in your dev env which causes issues that are really hard to understand.  If you have the version 0.5.2 of workgraph you do not know if this is from a local build (pip install . ) or from pypi (pip install aiida-workgraph). For example I got node_graph dependency issues and I did not understand why. This was because the pypi version of workgraph fixed the node_graph version to a different version than the current main version. This happened because I tried to reproduce the error locally using the pip freeze from RTD. To distinguish the current version on main from the pypi version one can add a post tag to the version so it is clear which version is used. 
- [ ] Fix the actual bug:
> There seems to be a really weird dependency resolvement happening when aiida-core is both installed through conda and pypi. The error does not really make sense, and I could not reproduce it locally. When you install aiida only from on source (conda or pypi) this does not happen.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,7 +8,7 @@ build:
   jobs:
     post_create_environment:
       - python -m pip install --no-cache-dir .[docs]
-      - pip list
+      - pip freeze
       - rabbitmq-server -detached
       - sleep 10
       - rabbitmq-diagnostics status
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -3,5 +3,6 @@ channels:
   - conda-forge
   - defaults
 dependencies:
+  - graphviz
   - aiida-core.services
   - qe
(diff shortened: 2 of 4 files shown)


superstar54:
Hi @agoscinski, thank you for investigating this.

> There seems to be a really weird dependency resolvement happening when aiida-core is both installed through conda and pypi.

Indeed, this is strange, and I have no idea too. Anyway, since in this PR #689  aiida-workgraph only supports aiida-core >=2.7, I removed aiida-core from the conda env.

Not sure, we will investigate this further, so I'm closing this PR.

PR #485, opened 2025-07-07
Revise RTD HowTos/graph_builder

GeigerJ2:
Top-level changes:
- Provide a clear list why one would use Graph Builder over just a normal task in the beginning, to make it more obvious what the feature is for
- Remove unnecessary `ctx` part to keep it simpler
- Minor text edits


docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -93,20 +92,27 @@ def add_multiply(x=None, y=None, z=None):
 
 from aiida_workgraph.utils import generate_node_graph
 
+# NOTE: Show only part of provenance graph? Like this is too big?
```
GeigerJ2:
```suggestion
# PRCOMMENT: Show only part of provenance graph? Like this is too big?
```

docs/gallery/howto/autogen/graph_builder.py
```diff
@@ -93,20 +92,27 @@ def add_multiply(x=None, y=None, z=None):
...
 # Graph builder
 # =============
 # A much more powerful tool to create nested WorkGraphs is the Graph Builder.
-# It is a decorator that we can add to a function that returns a WorkGraph
-# similar as `add_multiply` to have more control what we want to expose to the
-# user and to create dynamic workflows.
+# It is a decorator that we can add to a function that returns a WorkGraph.
+# The main advantages of using the Graph Builder are the following:
+# * **Clean interfaces** - Expose only necessary inputs/outputs while hiding internal complexity
+# * **Control flow** - Use Python `if`, `for`, `while` statements directly in workflows
```
GeigerJ2:
```suggestion
# * **Control flow** - Use Python `if` and `for` statements directly in workflows
```

superstar54:
Hi @GeigerJ2 , could you update the branch and fix the failed pre-commit test?

GeigerJ2:
Note: Add better documentation here on the difference between adding a WG as a task, and adding a `graph_builder`. This requires further discussion @superstar54 @edan-bainglass @agoscinski

edan-bainglass:
> Note: Add better documentation here on the difference between adding a WG as a task, and adding a `graph_builder`. This requires further discussion @superstar54 @edan-bainglass @agoscinski

@superstar54 and I were discussing that perhaps the graph builder should be properly introduced as a **concept**. Here in the HowTos, we would then reference graph builder where appropriate. Suggestions are workflow composition, authoring dynamic workflow, etc. TBD

GeigerJ2:
> @superstar54 and I were discussing that perhaps the graph builder should be properly introduced as a **concept**. Here in the HowTos, we would then reference graph builder where appropriate. Suggestions are workflow composition, authoring dynamic workflow, etc. TBD

Yeah, definitely something we can keep in mind. Overall, I feel like the HowTo section is a bit overwhelming with so many sections. Maybe we can crystallize out some general overarching concepts and present them nicer in _Concepts_. TBD

agoscinski (approved):

PR #498, opened 2025-07-07
Move note regarding `add_input` to after running the workgraph

edan-bainglass:
For consistency with the following section of the notebook.


superstar54 (approved):
LGTM!

PR #501, opened 2025-07-08
Rename active zones to `If`, `While`, and `Map`

GeigerJ2:


src/aiida_workgraph/decorator.py
```diff
@@ -85,6 +85,7 @@ def wrapper(*call_args, **call_kwargs):
         if graph is None:
             raise RuntimeError(f"No active Graph available for {func.__name__}.")
         task = graph.add_task(TaskCls)
+        # PRCOMMENT: Not sure if this now still works, as this uses `_active_zone`
         active_zone = getattr(graph, "_active_zone", None)
         if active_zone:
             active_zone.children.add(task)
```
GeigerJ2: @superstar54, this is still something to consider here

GeigerJ2: Not a problem

superstar54 (approved):
LGTM!

PR #503, opened 2025-07-08
Serialize the group_inputs and ctx

### LINKED ISSUE #493: Graph-level input serialization
In the following snippet

```python
wg = Workgraph()
wg.inputs = {"x": 1}
wg.add_task(add, name="add1", x=wg.inputs.x, y=2)
wg.add_task(add, name="add2", x=wg.inputs.x, y=3)
```

the two `add` tasks do not share the same `x` node, only the value. This is because it is not serialized at the input level, leading to a serialization of the `int` 1 into `orm.Int(1)` at the task level, **per task!**

To improve the provenance, it is best to serialize at the graph-level, i.e. `wg.inputs = {"x": orm.Int(1)}. One would consider automating this, such that the user can still provide `{"x": 1}`. However, auto-serialization can lead to issues when users provide non-serializable inputs.

superstar54:
Fix #493 
Serialize group inputs and the context variables when submitting the WorkGraph.

To be consistent, we use the same serializers as the ones used by `pyfunction` and `PythonJob`.

I also updated the docs. In the documentation, I mentioned a **Data Serialization** section, which does not yet exist. But it will be added soon.

### DIFF
--- a/docs/source/howto/graph_inputs_outputs.ipynb
+++ b/docs/source/howto/graph_inputs_outputs.ipynb
@@ -363,6 +363,28 @@
   },
   {
    "cell_type": "markdown",
+   "id": "70baac0a",
+   "metadata": {},
+   "source": [
+    "### Automatic serialization of raw Python data\n",
+    "\n",
+    "You can also pass raw Python values directly as inputs:\n",
+    "\n",
+    "```python\n",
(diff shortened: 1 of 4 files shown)


edan-bainglass (approved):
LGTM! Thanks for updating the docs as well 🙏

PR #504, opened 2025-07-08
Discard protocol docs

edan-bainglass:
The notebook was too specific for the use case of QE materials science. Dropping it in favor of keeping the HowTo sections general.


superstar54 (approved):
LGTM!

PR #510, opened 2025-07-09
Update map

superstar54:
This PR adds a new `MapItem` task inside the map zone instead of using the context variable to pass the item. This makes the GUI clear.

In the engine, we also copy this `map_item` task, and set the result of using the item. This also makes the logic in the engine clear.

```python
from aiida_workgraph import WorkGraph, task, Map
from aiida import orm, load_profile

load_profile()

@task
def add(x, y):
    return x+y

len_list = 2
data_x = {f"x_{i}": i for i in range(len_list)}
data_y = {f"y_{i}": -i for i in range(len_list)}

data = {""}

with WorkGraph("parallel_map") as wg:
    with Map(data_x) as map_zone_x:
        with Map(data_y) as map_zone_y:
            result = add(x=map_zone_x.item, y=map_zone_y.item)
    wg.run()
```

image


superstar54:
Hi @agoscinski ,I tested it locally, and it works for the nested map. I am going to merge this PR, so that our docs PR can use it.

Feel free to add comments, so we can improve in future PRs.

PR #517, opened 2025-07-09
Consolidate map zone and aggregate into parallel how-to

agoscinski:
This consolidates the how-to for aggregating data and mapping zones into the how-to on running tasks in parallel.

When I use `wg.submit` (which I used to reduce the verbose of `wg.run`) the notebook takes ~40s to run, while running with `wg.run()` it takes ~28s. When I was working on it locally it took actually 14s. I think this is because the RTD CI only offers 2 processes to run on, so not much parallelization. I can reduce the size of the list to 3 instead of 4, that should save probably another 4 seconds or so.

Preview: https://aiida-workgraph--517.org.readthedocs.build/en/517/howto/autogen/parallel.html


docs/gallery/howto/autogen/parallel.py
```diff
@@ -1,149 +1,226 @@
...
 # %%
-# First workflow
-# ==============
-# Suppose we want to calculate ```(x + y) * z ``` in two steps. First, add `x` and `y`, then multiply the result with `z`. And `X` is a list of values. We want to calculate these in parallel.
+# Perfectly parallelizable problem
+# ================================
+# A perfectly parallelizable problem can be broken down into smaller, independent subproblems that require no shared resources.
+# For example, consider an addition operation ``x + y`` applied element-wise to two lists: ``[x₁, ..., xₙ]`` and ``[y₁, ..., yₙ]``.
+# Each individual addition can be performed independently of the others.
+# ``WorkGraph`` automatically parallelizes task execution when there are no data dependencies between tasks (for more details on this concept, refer to `Dataflow programming <../../concept/autogen/dataflow_programming>`_).
```
agoscinski: this link does not exist yet. maybe reference in an issue?

agoscinski: change this to "WorkGraph Engine"

docs/gallery/howto/autogen/parallel.py
```diff
@@ -1,149 +1,226 @@
...
+# Provenance graph
+# ~~~~~~~~~~~~~~~~
 generate_node_graph(wg.pk)
 
 # %%
-# Second workflow: gather results
-# ================================
-# Now I want to gather the results from the previous `multiply_parallel` tasks and calculate the sum of all their results.
-# Let's update the `multiply_parallel` function to `multiply_parallel_gather`.
+# Map context
```
agoscinski: TODO change this to `Map context manager`

docs/gallery/howto/autogen/parallel.py
```diff
@@ -1,149 +1,226 @@
...
 """
+
 # %%
 # Introduction
 # ============
-# In this tutorial, you will learn how to run task in parallel.
-#
-# Load the AiiDA profile.
-#
+# Once you have developed a correct and functioning workflow, the next step is often to scale it up for large datasets.
```
superstar54, line 10: The phrase "scale it up for large datasets" may be misleading in this context. The use case to scale workflow to many independent data points (e.g., run relax for multiple structures), which would typically require submitting each workgraph for each data point. Because each data point is not related.

This how-to focuses on the use case: where users run multiple tasks within the same workflow (e.g., k-point convergence tests for a single structure with different k-points). Consider rephrasing to better reflect this.

superstar54 (approved):
LGTM!

Only one mirror comment on the use case, after fix this, please go ahead to merge the PR.

docs/gallery/howto/autogen/parallel.py
agoscinski: But your point is exactly addressed in the next sentence

> This typically involves applying the same workflow to many independent data points.

docs/gallery/howto/autogen/parallel.py
agoscinski:
```suggestion
# ``WorkGraph`` automatically parallelizes task execution when there are no data dependencies between tasks (for more details on this concept, refer to `WorkGraph Engine <../../concept/autogen/engine>`_).
```

docs/gallery/howto/autogen/parallel.py
agoscinski:
```suggestion
# Map context manager
```

docs/gallery/howto/autogen/parallel.py
agoscinski: Feel free to change this after the merge. We need to merge to reshuffle today later on

PR #524, opened 2025-07-10
Rework while how-to

agoscinski:
- adapt to new syntax
- make the context manager and task/programmatic approaches similar (i think workflow view was not the same before)

I think it needs another polish on the text but we can for now merge it since we want to any reshuffle the how-to's


docs/gallery/howto/autogen/while.py
```diff
@@ -0,0 +1,151 @@
...
+# %%
+# Provenance graph
+# ~~~~~~~~~~~~~~~~
+generate_node_graph(wg.pk)
+
+
+# %%
+# Further reading
+# ---------------
+# Similarly, other the control logic can implemented, see `Use if condition  <../if.ipynb>`_ for if logic and `How to run tasks in parallel <parallel.py>`_ for map logic.
```
agoscinski, line 151: double check if `parallel.py` works once #517 is merged

superstar54 (approved):
LGTM!

PR #542, opened 2025-07-11
Remove references to protocol section

agoscinski:
Note that a reference has to be also removed from `howto/transfer_workchain.ipynb` but I did not touch it because @superstar54  is responsible for moving this to some other higher-level section


superstar54 (approved):
LGTM!

PR #543, opened 2025-07-11
Move QE tutorials to notebooks to save CI time

GeigerJ2:
Preview:
https://aiida-workgraph--543.org.readthedocs.build/en/543/


GeigerJ2:
We decided not to do this for now, as it doesn't save considerable time

PR #544, opened 2025-07-11
Remove qe dependency in docs and move tutorial to Jupyter notebooks

agoscinski:
Remove the qe dependency in RTD and convert the tutorials that rely on quantum espresso to Jupyter notebooks.

I prioritized this to speed up the RTD build for everyone. I downloaded the notebook version you can get from RTD at the moment. I did not execute because there is no qe for macos. Maybe we can do this in future PR to get the RTD build speedups? @superstar54 what you think.

Side note for future PR: I think `docs/requirements.txt` is not needed anymore


agoscinski:
decided to not do it

Issue #521, opened 2025-07-10
CI: Move to the `aiida-core` tech stack, e.g. Ruff


GeigerJ2:

superstar54:
Hi @GeigerJ2, I believe this is now complete.

PR #554, opened 2025-07-12
Refactor `@task.graph_builder ` to `@task.graph`

superstar54:
Recently, we found that the `graph_builder` is difficult touse in the context manager approach, for example:
- how to call the `graph_buider` as a task, i.e., as a black box

This is because the `graph_builder` returns a WorkGraph, while in the context manager approach, one calls every component as a task and expects it to return the outputs. Even though `graph_builder` is successfully used in the node-graph programming approach, it is created before the context manager approach. Thus, in this PR, we refactor it to make it work more consistently inside the context manager approach, while still usable in the node-graph programming approach.

Here is an example for the new syntax:
```python

@task()
def add(x, y):
    return x + y

@task()
def multiply(x, y):
    return x * y

@task.graph()
def add_multiply(x, y, z):
    outputs1 = add(x=x, y=y)
    outputs2 = multiply(x=z, y=outputs1.result)
    return outputs2.result
```

Use it in inside workflow:
```python
with WorkGraph() as main_wg:
    outputs = add(x=10, y=5)
    # Use the Graph Builder as a task, feeding it the future result.
    add_multiply(x=outputs.add_result, y=2, z=3)
```

Get the WorkGraph directly from a graph task:
```python
wg = add_multiply.get_graph(1, 2, 3)
```


src/aiida_workgraph/decorator.py
```diff
@@ -3,6 +3,7 @@
 from typing import Any, Callable, Dict, List, Optional, Union, Tuple
 from aiida.engine import calcfunction, workfunction, CalcJob, WorkChain
 from aiida_workgraph.task import Task
+from .workgraph import WorkGraph
```
edan-bainglass, line 6: Need to sort imports. Don't worry about it now. When we switch to Ruff (#521), this will be automated.

edan-bainglass (approved):
LGTM! I'll know better after I use for a bit 😅

superstar54:
> LGTM! I'll know better after I use for a bit

OK!

@agoscinski , @GeigerJ2 , since the context manager approach is the main direction we want to take, I am going to merge this PR now. We can address any bugs that are found during the reconstruction of the docs.

src/aiida_workgraph/decorator.py
superstar54: Agree

PR #559, opened 2025-07-13
Update graph-level inputs-outputs section

edan-bainglass:
Updating section with context manager syntax.


edan-bainglass:
@superstar54 please have a look at the **Providing graph-level inputs metadata** sub-section. I made a comment regarding the impact on the GUI when providing an identifier to a graph-level input. I'd like to show this impact visually. How can I do this?

PR #562, opened 2025-07-13
Merge save and restart into continue a workflow section

edan-bainglass:
This PR updates the continue a workflow section with the new syntax and merges any relevant docs from the save and restart section, which is discarded.


docs/source/howto/continue_finished_workgraph.ipynb
```diff
@@ -72,89 +63,178 @@
...
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "with WorkGraph.load(wg.pk) as wg2:\n",
+    "    wg2.name = \"AddMultiplyContinued\"\n",
+    "    wg2.inputs = dict.fromkeys(['x'])\n",
+    "    wg2.restart()\n",
+    "    sum_ = wg2.outputs.result + wg2.inputs.x\n",
```
superstar54: the graph-level outputs, `wg.outputs.result`, is not recommended as the input of another task, ortherwise, it may create a cyclic loop, e.g.,
```python
wg2.outputs.result = wg2.outputs.result + wg2.inputs.x
```

superstar54 (changes requested):
Continue a finished workflow includes two use cases:

- append new tasks, and run (covered by this doc)
- modify the inputs of the old task, this will reset all child tasks of the modified tasks.

The second user case was in the old `save and restart`. I would suggest adding it to here too.

superstar54 (approved):
LGTM!

PR #568, opened 2025-07-13
Fix pickle error for AiiDA process

### LINKED ISSUE #564: Error when wrapping a CalcJob as a task
```python
from aiida.calculations.arithmetic.add import ArithmeticAddCalculation

load_profile()

AddTask = task()(ArithmeticAddCalculation)



code = orm.load_code("add@localhost")

@task.graph()
def add_workgraph(x, y, code):
    
    outputs1 = AddTask(x=x, y=y, code=code)
    outputs2 = AddTask(x=x, y=outputs1.sum
                   , code=code)
    return outputs2.sum

wg = add_workgraph.get_graph(x=1, y=2, code=code)
wg.run()

print("sum: ", wg.outputs.result.value)
```
raise an error:
```
$ python test_pickle.py
Traceback (most recent call last):
File "/home/xing/tests/workgraph/migration/test_pickle.py", line 24, in <module>
pickle.dump(AddTask, open('task.pkl', 'wb'))
File "/home/xing/apps/miniforge3/envs/aiida/lib/python3.11/site-packages/cloudpickle/cloudpickle.py", line 1511, in dump
Pickler(file, protocol=protocol, buffer_callback=buffer_callback).dump(obj)
File "/home/xing/apps/miniforge3/envs/aiida/lib/python3.11/site-packages/cloudpickle/cloudpickle.py", line 1295, in dump
return super().dump(obj)
^^^^^^^^^^^^^^^^^
TypeError: cannot pickle '_abc._abc_data' object
```

superstar54:
Fix #564 

Cloudpickle tries to serialize the decorator wrapper function (`_make_wrapper`), its closure and attributes. The AiiDA process (CalcJob), which inherits the `ABCMeta`  is inside the wrapper function. Thus, we get this 

```
TypeError: cannot pickle '_abc._abc_data' object
```

Actually, only when the callable is a function do we need to pass it to the wrapper function to inspect the function signature and prepare the inputs correctly.

If the callable is a CalcJob or WorkChain, we don't need to pass it to the wrapper function.

This PR solves the issue by checking if the callable is a function before passing it to the wrapper function. If it is not a function, we simply pass None to the wrapper function.


src/aiida_workgraph/decorator.py
```diff
@@ -262,7 +264,7 @@ def decorator_graph(
         """
 
         def decorator(func):
-            from aiida_workgraph.tasks.builtins import GraphBuilderTask
+            from aiida_workgraph.tasks.builtins import GraphTask
```
edan-bainglass, line 267: How is this related (and the rest of the touchups w.r.t the graph builder)?

edan-bainglass (approved):
LGTM! One comment regarding graph builder change

src/aiida_workgraph/decorator.py
superstar54: Good catch! The class name change is not directly related to the fix in this PR, but is a leftover from a previous PR.  And I noticed it while I was working on the same file.

PR #569, opened 2025-07-13
Add default interval and timeout inputs for monitor task

### LINKED ISSUE #567: Monitor Task no longer accepts `interval` and `timeout`
For example

```python
import os


@task.monitor
def file_monitor(filepath):
    return os.path.exists(filepath)


with WorkGraph("FileMonitor") as wg:
    wg.inputs = {"monitor": dict.fromkeys(["filepath", "interval", "timeout"])}
    file_monitor(
        filepath=wg.inputs.monitor.filepath,
        interval=wg.inputs.monitor.interval,
        timeout=wg.inputs.monitor.timeout,
    ) >> sleep_add(x=1, y=2)

wg.run(
    inputs={
        "graph_inputs": {
            "monitor": {
                "filepath": "/tmp/monitor_test.txt",
                "interval": 1,
                "timeout": 10,
            }
        }
    }
)
```
yields `Exception: No property named interval. Accept name are ['filepath', '_wait']`

@superstar54

superstar54:
- Add default interval and timeout inputs for monitor task
- Check if funciton has any forbidden inputs: `interval`, `timeout`
- Separate tests for `monitor` and `awaitable`.


src/aiida_workgraph/decorator.py
```diff
@@ -382,9 +382,33 @@ def decorator(func):
...
 
+            sig = inspect.signature(func)
+            forbidden = {"interval", "timeout"}
+            conflicts = forbidden.intersection(sig.parameters)
+            if conflicts:
+                raise ValueError(
+                    f"Function '{func.__name__}' defines parameter(s) {sorted(conflicts)}, "
+                    "which conflict with default 'monitor' arguments."
+                )
+            task_inputs = inputs if inputs is not None else []
```
edan-bainglass, line 393:
```suggestion
            task_inputs = inputs or []
```

src/aiida_workgraph/tasks/factory/awaitable_task.py
```diff
@@ -41,6 +41,9 @@ def execute(self, engine_process, args=None, kwargs=None, var_kwargs=None):
         from aiida_workgraph.executors.monitors import monitor
 
         executor = NodeExecutor(**self.get_executor()).executor
+        # get the raw function without the decorator
+        if hasattr(executor, "_func"):
+            executor = executor._func
```
edan-bainglass, line 46: What's this about?

edan-bainglass (approved):
LGTM! Good idea to refactor the tests 👍 Just a few comments, then you're good to go 🙂

src/aiida_workgraph/tasks/factory/awaitable_task.py
superstar54:
```
@task 
def add(x, y):
    return x + y
```
In the above example, the raw `add` function is stored as `_func` attribute of the decorated function.

I added an `if` check, in case the function is not decorated, but is used as a monitor directly somehow (which I don't remember how).

src/aiida_workgraph/decorator.py
superstar54: Yes, `inputs or []` is simpler, which I wrote before. But there is still a little bit difference. If the inputs is a empty `[]`, the first case will use the orginal `[]`, while the second case will use a new `[]`.
so I keep the longer one, just to make sure the user's orginal inputs is passed even when it is a empty list.

PR #570, opened 2025-07-14
Fix serialization failure of the outputs in Widget view

### LINKED ISSUE #560: `WorkGraph.restart()` should reset the state
Currently, the state of the completed workgraph remains `FINISHED`. Should be `PLANNED`.

@superstar54

### LINKED ISSUE #561: Completed task serialization failure
After
```python
with WorkGraph("AddMultiplyToBeContinued") as wg:
    wg.inputs = dict.fromkeys(['x', 'y', "z"])
    wg.outputs.result = (wg.inputs.x + wg.inputs.y) * wg.inputs.z
    
wg.submit(
    inputs={
        "graph_inputs": {
            "x": 1,
            "y": 2,
            "z": 3
        }
    },
    wait=True,
)
```
doing
```python
with WorkGraph.load(wg.pk) as wg2:
    wg2.restart()
    sum_ = wg2.outputs.result + 5
```

and trying to display `wg2.tasks.op_add` results in
```python
ValueError: Can't clean for JSON: <Int: uuid: 2754e1f3-c1eb-4548-9e42-a3d0a22a10b0 (pk: 10827) value: 3>
```

@superstar54

superstar54:
Fix #561, At the same time, fix #560  

- Use the same helper function from the `WorkGraph`,  `workgraph_to_short_json`, to serialize the data of the task. In this way, we keep the API consistent. 
- But in the default view of WorkGrpah, only the required inputs are shown, and the connected outputs are shown. While in the task view, we want to show all the inputs and outputs. But, in principle, all the logic to control which inputs/outputs to be shown should be moved to the frontend, so there is an option to let the user control on the GUI directly. This will be do in a future PR in the Widget side.


src/aiida_workgraph/workgraph.py
```diff
@@ -524,6 +524,7 @@ def restart(self):
         # so that the WorkGraphSaver can compare the difference, and reset the modified tasks
         self.restart_process = self.process
         self.process = None
+        self.state = "PLANNED"
```
edan-bainglass, line 527: Can we do this in a separate PR?

edan-bainglass (approved):
LGTM! See my comment about the restart

src/aiida_workgraph/workgraph.py
superstar54: I am too lazy to open a PR to fix one line of code. :rofl: 
Will do it next time.

PR #571, opened 2025-07-14
Set default name for all AiiDA component tasks

superstar54:
```python
from aiida_quantumespresso.workflows.pw.relax import PwRelaxWorkChain
PwRelaxTask = task()(PwRelaxWorkChain)
```
Then show the `PwRelaxTask`. Before the name is `workchain`:
Screenshot from 2025-07-14 10-55-50

After this PR, the name will be `PwRelaxWorkChain`:
Screenshot from 2025-07-14 10-54-58

The `AUTO_CALCFUNCTION` option does not exist, thus is deleted.


edan-bainglass (approved):
LGTM!

PR #575, opened 2025-07-15
Convert how-to sections to gallery scripts

edan-bainglass:


docs/gallery/howto/autogen/combine_workgraphs.py
```diff
@@ -0,0 +1,119 @@
...
+generate_node_graph(wg2.pk)
+
+# %%
+# Making it reusable
+# -------------------
+#
+# Now, how do we make this reusable? Simple! We just wrap **AddMultiply** in a generating function.
+
+
+def generate_add_multiply_workgraph():
```
superstar54, line 103: Do you want to change it to a `@task.graph` or in a future PR? Please open an issue for the latter cases.

docs/gallery/howto/autogen/async.py
```diff
@@ -0,0 +1,68 @@
+"""
+Run tasks asynchronously
```
superstar54: All tasks in WorkGraph are run asynchronously, so maybe change to another title, e.g., `Run async functions as Tasks`

docs/gallery/howto/autogen/async.py
```diff
@@ -0,0 +1,68 @@
...
+# %%
+import asyncio
+
+from aiida import load_profile
+from aiida_workgraph import WorkGraph, task
+
+load_profile()
+
+
+# %%
```
superstar54, line 28:
```suggestion
# %%
# Use the ``@task.awaitable`` decorator on an async def function to make it non‑blocking:
#
```

docs/gallery/howto/autogen/monitor.py
```diff
@@ -0,0 +1,199 @@
+"""
+Assign a monitoring task
```
superstar54: `monitoring task` is not clear. Consider changing the title to e.g., `Monitor external events with polling.`

docs/source/howto/index.rst
```diff
@@ -9,19 +9,19 @@ This section contains a collection of HowTos for various topics.
    :caption: Contents:
 
    autogen/use_calcjob_workchain
-   graph_inputs_outputs
+   autogen/graph_inputs_outputs
```
superstar54: The file name `graph_inputs_outputs` does not match, so the doc is not shown.

superstar54 (changes requested):
The report works, nice! 
I added a few change requests.

docs/gallery/howto/autogen/combine_workgraphs.py
edan-bainglass: We need to discuss this. Since this works, it may create confusion for the user, e.g., why use `@task.graph` if I can just do this?

docs/gallery/howto/autogen/monitor.py
edan-bainglass: How about "Monitor external events as a task"?

superstar54: Good!

docs/gallery/howto/autogen/combine_workgraphs.py
edan-bainglass: I'll open an issue

PR #576, opened 2025-07-15
Merge graph-level tasks to the normal tasks

### LINKED ISSUE #574: Modifications to graph-level inputs aren't detected
@superstar54

superstar54:
Fix 574.

Adapt the changes in `node-graph`, this PR: https://github.com/scinode/node-graph/pull/66.

The `graph_inputs`, `graph_outputs` and `graph_ctx` tasks are in the same Task collection as other tasks. In the Engine, at the setup stage, these tasks are set to be `FINISHED`, so that tasks that have input links to this task do not need to wait.


superstar54:
@edan-bainglass , with this PR. The restart example using graph-level inputs will work.

```python
with WorkGraph("restart") as wg:
    wg.inputs = {"x": 1, "y": 2, "z": 3}
    wg.outputs.result = (wg.inputs.x + wg.inputs.y) * wg.inputs.z
    
wg.run()
print("Result:", wg.outputs.result.value)

with WorkGraph.load(wg.pk) as wg2:
    wg2.restart()
    wg2.inputs.x = 2

wg2.run()
print("Result:", wg2.outputs.result.value)
```

src/aiida_workgraph/tasks/builtins.py
```diff
@@ -1,5 +1,52 @@
...
+
+
+class GraphOutputs(GraphLevelTask):
+    identifier = "workgraph.graph_outputs"
+    name = "Graph_Outputs"
+
+
+class GraphContext(GraphLevelTask):
+    identifier = "workgraph.graph_ctx"
+    name = "Graph_Ctx"
```
edan-bainglass, line 49: Are these redefinitions of `node-graph` classes?

edan-bainglass (approved):
LGTM! Just a question about some classes that look like redefinitions from `node-graph`

src/aiida_workgraph/tasks/builtins.py
edan-bainglass: Maybe redefinition is not quite right. But still, very similar to [this](https://github.com/scinode/node-graph/blob/8f4b3020614b018cdd7cbd0e6cfa24b66759e3be/src/node_graph/nodes/builtins.py). Can you explain a bit? 🙏

superstar54: Yes, this is the redefinition of the classes. `aiida-workgraph` is built on top of `node-graph`, but they use different plugin entry point name, thus WorkGraph only see the tasks from this group.
```
[project.entry-points."aiida_workgraph.task"]
```
That's why we redefine the `GraphInputs`, `GraphOutputs` and `GraphCtx` classes.
But, for the `GraphLevelTask`, it would be good to inherit from the `GraphLevelNode` from the node-graph directly.
However, due to the dynamic feature (the inputs of these Tasks are dynamic changed), these classes are created from a Class Factory that is aware of the entry point group.
To be short, I haven't found a good way to inherit the class :rofl: .

PR #577, opened 2025-07-15
[Docs] add remote_job

superstar54:
Add a new section: **Run calculation remotely**

Preview: https://aiida-workgraph--577.org.readthedocs.build/en/577/howto/autogen/remote_job.html


docs/gallery/howto/autogen/remote_job.py
```diff
@@ -0,0 +1,94 @@
...
+===============================
+
+"""
+
+# %%
+# Introduction
+# ============
+# There are three ways to run calculations remotely using AiiDA:
+#
+# - `CalcJob`: run a shell command with a dedicated AiiDA plugin.
```
edan-bainglass: This is not very clear. What does aiida-core docs say?

docs/gallery/howto/autogen/remote_job.py
```diff
@@ -0,0 +1,94 @@
...
+# - `ShellJob`: run any shell command on a remote computer.
+# - `PythonJob`: run any Python function on a remote computer.
+#
+# For `CalcJob`, please refer to `How to use aiida-core components inside WorkGraph <../../howto/autogen/use_calcjob_workchain.html>`_.
+# For `ShellJob`, please refer to the `Run shell commands as a task <../../howto/autogen/shelljob.html>`_.
+# In this tutorial, we will focus on `PythonJob`.
+
+# First, load the AiiDA profile.
```
edan-bainglass: Remove this line. We don't need to explicitly say anything.

docs/gallery/howto/autogen/remote_job.py
```diff
@@ -0,0 +1,94 @@
...
+
+# Decorator to define a pythonjob
+@task.pythonjob()
+def add(x, y):
+    return x + y
+
+
+with WorkGraph() as wg:
+    # Call the add task, specifying 'localhost' as the computer.
+    outputs = add(x=1, y=2, computer="localhost")
```
edan-bainglass, line 43: IDEs will complain unfortunately, as add only takes x and y. But fine for now

docs/gallery/howto/autogen/remote_job.py
```diff
@@ -0,0 +1,94 @@
...
+    wg.run()
```
edan-bainglass, line 44:
```suggestion
wg.run()
```

docs/gallery/howto/autogen/remote_job.py
```diff
@@ -0,0 +1,94 @@
...
+
+# First, load the AiiDA profile.
+from aiida import load_profile
+
+load_profile()
+
+# %%
+# PythonJob task
+# ========================
+# One can create a `PythonJob` task using the `@task.pythonjob()` decorator.
```
edan-bainglass:
```suggestion
# One can create a ``PythonJob`` task using the ``@task.pythonjob()`` decorator.
```

edan-bainglass: RST needs double ticks to display as code

docs/gallery/howto/autogen/remote_job.py
edan-bainglass: Same throughout. Shouldn't be in the context manager.

docs/gallery/howto/autogen/remote_job.py
```diff
@@ -0,0 +1,94 @@
...
+    wg.run()
+
+# Print out the result once the WorkGraph has finished executing.
+print("\nResult: ", outputs.result)
+
+# %%
+# Prepare Python environment on remote computer
+# -----------------------------------------------
+# `PythonJob` requires that the Python version on the remote computer matches the one used
+# on the localhost where AiiDA is running. You can use `conda` to create a virtual
```
edan-bainglass: Try one sentence per line in the script. Easier to maintain 🙏

docs/gallery/howto/autogen/remote_job.py
```diff
@@ -0,0 +1,94 @@
...
+# environment with the same Python version.
+# Then, activate the environment in the `metadata` of the `PythonJob` task.
```
edan-bainglass:
```suggestion
# environment with the same Python version, then activate the environment in the `metadata` of the `PythonJob` task.
```

edan-bainglass (changes requested):
Overall good. See comments 🙏

docs/gallery/howto/autogen/remote_job.py
```diff
@@ -0,0 +1,94 @@
+"""
+
+===============================
+Run calculation remotely
```
edan-bainglass:
```suggestion
Run calculations remotely
```

docs/gallery/howto/autogen/remote_job.py
superstar54: Actually, based on the typical usage of context managers, `wg.run()` should be inside the with block. Like the 
```python
with open(file) as f:
  f...
  # all other operations are inside
```
Because when the with block exits, it usually runs some cleanup. Currently, WorkGraph does not run any cleanup when exiting the block. Therefore, inside and outside are the same. However, inside is more typical usage.

docs/gallery/howto/autogen/remote_job.py
superstar54: I have added a paragraph to explain the additional inputs and outputs for the users.

docs/gallery/howto/autogen/remote_job.py
superstar54: I change it to 
```
# - `CalcJob`: run a shell command (via dedicated AiiDA plugins to manage inputs, outputs) on a remote computer.
#    Please refer to `How to use aiida-core components inside WorkGraph <../../howto/autogen/use_calcjob_workchain.html>`_.
```

docs/gallery/howto/autogen/remote_job.py
edan-bainglass: Not always the case. In Solara, for example, you can define the children of a component via a context manager:
```python
with solara.Div() as section:
    solara.Text("Hello")
    solara.Text("World")
return section
```
Note that I often in the docs use the context manager to construct `wg`, then make some comments or print it out with `wg.to_html()`, and only then do `wg.submit()` or `wg.run()` outside of the block.

docs/gallery/howto/autogen/remote_job.py
edan-bainglass: My issue is specifically with "run a shell command". Didn't seem clear/accurate to me. In AiiDA docs, we say the following:
```
However, not all computations are well suited to be implemented as a python function, 
but rather are implemented as a separate code, external to AiiDA. 
To interface an external code with the engine of AiiDA, the CalcJob process class was introduced.
```
This description doesn't immediately tell me that I'm wrapping a shell command. Maybe an executable is more accurate? 
From chatGPT:
```
In AiiDA, a CalcJob is a specific type of calculation process designed to run an external code 
(such as a DFT software, molecular dynamics simulation, or custom program) on a computational resource 
(computer or HPC cluster)
```
and
```
It’s a subclass of Process in AiiDA’s engine, tailored for launching, tracking, 
and retrieving execution of external binaries.
```

But I guess an executable/binary is a "shell command". I'll let you decide.

docs/gallery/howto/autogen/remote_job.py
superstar54: If we want to show the GUI of wg in the Jupyter notebooks, then yes, we have to move the `wg` or `wg.to_html()` outside the with block.

docs/gallery/howto/autogen/remote_job.py
superstar54: Yes, it's not easy to explain the `CalcJob` in one sentence, but since we ask the user to refer to the `How to use aiida-core components inside WorkGraph` section, so they will learn the concepts there. So I think the current text is fine.

edan-bainglass (approved):
LGTM!

PR #578, opened 2025-07-15
Clean the files left over from the Scheduler

superstar54:
These functions are moved to the `aidia-scheduler`.


edan-bainglass (approved):
LGTM!

PR #579, opened 2025-07-15
Remove `While` WorkGraph in the engine

superstar54:
The `While` WorkGraph code is outdated. Also comment out the line to print out the `task results`.


edan-bainglass (approved):
LGTM!

PR #580, opened 2025-07-16
Allow graph-level inputs directly

edan-bainglass:
Now `wg.run(inputs={"x": 1, "y": 2})` instead of `wg.run(inputs={"graph_inputs": {"x": 1, "y": 2}})`

### DIFF
--- a/docs/gallery/howto/autogen/async.py
+++ b/docs/gallery/howto/autogen/async.py
@@ -42,10 +42,8 @@ with WorkGraph("AwaitableGraph") as wg:
 
 wg.run(
     inputs={
-        "graph_inputs": {
-            "x": 1,
-            "y": 2,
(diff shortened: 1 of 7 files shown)


superstar54 (changes requested):
In the `run` and `submit` methods, the code is duplicated. Could you add a method to set the inputs, e.g., `set_inputs`. And add the tests too.

edan-bainglass:
> In the `run` and `submit` methods, the code is duplicated. Could you add a method to set the inputs, e.g., `set_inputs`. And add the tests too.

Sorry, prematurely asked for review. I need to update the tests, but also the docs to follow this new api. But first, let's get #575 merged.

superstar54 (approved):
LGTM!

PR #581, opened 2025-07-16
Remove usage of aiida-scheduler

superstar54:
Since `aiida-scheduler` package has not made a stable release yet, the package name and API could change in the near future. Therefore, it's not good to add it as a dependency, considering that we are aiming at a stable release of aidia-workgraph.


src/aiida_workgraph/engine/workgraph.py
```diff
@@ -347,22 +347,6 @@ def message_receive(
...
-        """Submit process for execution.
-
-        :param process: The process class.
-        :param inputs: The dictionary of process inputs.
-        :return: The process node.
-        """
-        from aiida_scheduler.control import submit_to_scheduler_inside_workchain
-
-        return submit_to_scheduler_inside_workchain(self, process, inputs, **kwargs)
-
```
GeigerJ2, line 365: This is not used anywhere? At least from my search I didn't find it. Good if you can double-check, @superstar54.

GeigerJ2 (approved):

src/aiida_workgraph/engine/workgraph.py
superstar54: The code overrides the default `submit` method of the `Process` class. The method is used if you want to submit other process inside a process.

PR #585, opened 2025-07-16
Remove outdated builtin sections (monitor)

superstar54:
The previous builtin section includes PythonJob, ShellJob and Monitor tasks. All of them are moved to Howto section. This PR
- remove the builtin folder and related files
- also update the conf and index file


docs/source/built-in/index.rst
```diff
@@ -1,13 +0,0 @@
...
-===========================================
-
-This section contains a collection of built-in tasks that are used in the AiiDA WorkGraph.
-
-.. toctree::
-   :maxdepth: 1
-   :caption: Contents:
-
-   pythonjob
-   autogen/shelljob
```
GeigerJ2, line 12: Wondering why docs build wasn't failing if both of these were wrong references.

GeigerJ2 (approved):
I suggest to change the title to something like "Remove oudated builtin sections (monitor)", because that is the only actual section that is removed, and, for future reference, _if_ we ever need to find again where the `monitor` tutorial code went 👀

docs/source/built-in/index.rst
superstar54: I have the same question :laughing:

superstar54:
Agree, I changed the title.

PR #586, opened 2025-07-17
Update to `node-graph` `v0.2.15` and expose `group` utility

### LINKED ISSUE #582: Expose useful / necessary objects from node-graph via workgraph
For instance, `from node_graph.group import group`, should instead be importable from `aiida-workgraph`. I don't think we want to actually expose the user to the `node-graph` dependency in the docs, at all. Any other things that come to mind, @superstar54?

edan-bainglass:
Closes #582


superstar54 (approved):
LGTM!

PR #587, opened 2025-07-19
[Docs] add `Graph` task for While logic

superstar54:
The interesting thing is that the current `Graph` task supports the while logic out of the box. Surprise, but nice. :rocket: 

Here I update the doc, and a small change in the `graph` task api.

- In the doc `control-flow`, add `Graph` task for the `While` logic
- Update the decorator to allow the `graph` task to return a value too.
- In the `Graph` task execution, use a temporary workaround to fix the pickle error for the recursive call.

doc preview: https://aiida-workgraph--587.org.readthedocs.build/en/587/howto/autogen/control-flow.html#id7

### DIFF
--- a/docs/gallery/howto/autogen/control-flow.py
+++ b/docs/gallery/howto/autogen/control-flow.py
@@ -294,6 +294,42 @@ wg.to_html()
 
 generate_node_graph(wg.pk)
 
+# %%
+# Graph Task
+# ----------
+# We can also implement the same workflow using a *graph task* with the ``@task.graph`` decorator.
+# Let's implement the same logic as above, but this time using the `Graph Task` approach.
+
+
+# Wrap the function in a class to avoid pickling issues with the recursive call
+# This works because cls.sum_to_n provides a persistent reference within the instance.
+class MySummer:
+    @classmethod
+    @task.graph()
+    def add_multiply_while(cls, n, N):
+        if n < 8:
+            n = add(x=n, y=1).result
+            n = multiply(x=n, y=2).result
+            # Recursive call to continue the loop
+            return cls.add_multiply_while(n=n, N=N).result
+        else:
+            return n
+
+
+with WorkGraph("while_graph_task") as wg:
+    first_add_result = add(x=1, y=1).result
+    add_multiply_while_result = MySummer.add_multiply_while(
+        n=first_add_result, N=8
+    ).result
+    final_add_result = add(x=add_multiply_while_result, y=1).result
+    wg.outputs.result = final_add_result
+    wg.run()
+    print(f"Result            : {wg.outputs.result.value}")
+    assert wg.outputs.result.value == 15
+
+
+#%%
+# In this example, the key idea is the recursive call to the `add_multiply_while` method.
 
 # %%
 # Conclusion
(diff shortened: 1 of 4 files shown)


docs/gallery/howto/autogen/control-flow.py
```diff
@@ -294,6 +297,56 @@ def compare_lt(x, y):
 
 generate_node_graph(wg.pk)
 
+# %%
+# Graph Task
+# ----------
+# We can also implement the same while loop logic using a graph task with the @task.graph decorator.
```
edan-bainglass:
```suggestion
# We can also implement the same while loop logic using a graph task with the ``@task.graph`` decorator.
```

docs/gallery/howto/autogen/control-flow.py
```diff
@@ -294,6 +297,56 @@ def compare_lt(x, y):
...
+# Instead of a native while loop, this approach uses recursion: the graph task calls itself repeatedly until a termination condition is met.
```
edan-bainglass:
```suggestion
# Instead of a native ``while`` loop, this approach uses recursion: the graph task calls itself repeatedly until a termination condition is met.
```

docs/gallery/howto/autogen/control-flow.py
```diff
@@ -294,6 +297,56 @@ def compare_lt(x, y):
...
+    assert wg.outputs.result.value == 15
+
+
+#%%
+# This approach encapsulates the entire loop within a single, reusable task.
+
+# %%
+# Workflow view
+# ~~~~~~~~~~~~~
+# In the workflow view, add_multiply_while appears as a single "black box" task.
```
edan-bainglass, line 339: This is something that we need to discuss. I don't think it is blocking at the moment, but the fact that this approach does not lead to a proper while zone is uncomfortable to say the least.

docs/gallery/howto/autogen/control-flow.py
```diff
@@ -307,8 +360,11 @@ def compare_lt(x, y):
...
 #
-# - **While loops** use the ``While`` context manager to create iterative workflows with configurable
-#   maximum iterations to prevent infinite loops
+# - **While loops** use:
+#
+#   - The ``While`` context manager to create iterative workflows with configurable maximum iterations to prevent infinite loops
+#   - The ``@task.graph`` decorator, where loops are created through recursion to handle dynamic iterations.
+#
 # - **Context variables** (``wg.ctx``) require explicit dependency management using ``<<`` and ``>>`` operators
 #   since WorkGraph cannot automatically infer dependencies to avoid potential cycles
```
edan-bainglass:
```suggestion
#   since ``WorkGraph`` cannot automatically infer dependencies to avoid potential cycles
```

edan-bainglass (approved):
In general, LGTM! I am a bit concerned regarding the impact on the graph visualization. Also, what is the plan to remove the pickling workaround?

superstar54:
> I am a bit concerned regarding the impact on the graph visualization.

`Graph` task is always shown as a black-box before running, as soon as the task runs, one can inspect it. In the case of this recursive call, the GUI will show a deeply nested workflow. (The number of nested levels is determined by the number of recursive call).

>  Also, what is the plan to remove the pickling workaround?

I don't know how to replace the workaround for the moment, :rofl:

PR #591, opened 2025-07-20
Update quick start docs

edan-bainglass:


superstar54:
Thanks for the work! The `Quick Start` has a nice format with many useful tips and notes, but it looks a bit long and covers too many advanced concepts.

During the coding week, we concluded that `quick start` should only provide a very quick/simple example; thus, I think it only needs to introduce:

- Define tasks using decorator, `add` and `multiply`
- Create an `add_and_multiply` workflow using a context manager approach.
- Show the GUI, and results
- And at the end, provide a table with various links to different sections.

We will provide an `Overview` page, which introduces all the features (only showing features, not running code).

What do you think? @edan-bainglass  @GeigerJ2

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,415 @@
...
 #
-# If you haven't configured an AiiDA profile, you can run the following
+# This will also install ``aiida-core`` and its dependencies.
+
+# %%
+# Setup
+# -----
 #
-# .. code:: console
+# Logging
```
superstar54: Switching the log level to `REPORT` seems not essential here.

edan-bainglass:
I tried to touch on the top points (operator tasks, functional tasks, graph tasks, control flow). The reason I include the last one is to  emphasize to the user the importance to thinking in tasks.

Happy to discuss 🙂 As for the context manager up front, I'm no longer sure that's the best approach, having played with the graph decorator. But again, happy to discuss.

edan-bainglass:
@superstar54 I'm having this tested locally by someone to get an outside take on readability, clarity, length, etc.

But considering the refactoring we've been discussing, putting `@task.graph` up front and gathering all other approaches to their own blog-like sections (or categories with subsections), parts of the quick start will likely reduce to 1-2 sentences + reference.

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,380 @@
...
+#   - tasks to encapsulate a process
+#   - sub-workflows as tasks to encapsulate several sub-processes
 #
-#    pip install aiida-workgraph
+# - an example demonstrating control flow in ``WorkGraph``
 #
-# If you haven't configured an AiiDA profile, you can run the following
+# .. note::
+#
+#    This quick start tutorial assumes basic AiiDA knowledge.
```
superstar54: I would not assume this; otherwise, the users go to the AiiDA doc and may get lost there, or take a `long` time to get AiiDA basics (calcfunction, calcjob), but they are not used here.,

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,380 @@
...
+#
+# .. tip::
+#
+#    It is best to follow the quick start tutorial using a Jupyter notebook.
+
+# %%
+# Installation
+# ------------
+#
+# Let's first install ``aiida-workgraph`` (and ``aiida-core`` by extension):
```
superstar54: `by extension` is not clear for many users.
```suggestion
# Let's first install ``aiida-workgraph`` (this will also install ``aiida-core`` automatically):
```

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,380 @@
...
+# Let's run it with some inputs:
 
-from aiida_workgraph import task
+wg.run(inputs={"x": 2, "y": 3, "z": 4})
 
-# define add task
-@task()
-def add(x, y):
-    return x + y
+# %%
```
superstar54: I would avoid mentioning  `message broker`  and submit. It's not essential to run this quick start.

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,380 @@
...
+#
+#    Some features of ``WorkGraph`` are best demonstrated interactively.
+#    We recommend using the AiiDA GUI for this.
+#    You can learn how to run and use the GUI in the :doc:`../gui/web` section.
 
-######################################################################
-# First workflow
-# --------------
+# %%
+# Simple workflow
```
superstar54, line 50: In this section, socket operations are used to create the `add` and `multiply` tasks implicitly. However, socket operations are a relatively advanced concept and not the primary way to define tasks. I suggest removing this part and introducing the `@task` decorator approach directly, as it provides a much clearer and more intuitive way to define tasks and graphs.

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,380 @@
...
-# ---------------------
+# %%
+# Compare the graph above with the one we defined earlier.
+# Other than the task names, they are identical, as to be expected.
 #
-# One can use AiiDA components (``CalcJob`` and ``WorkChain``)
-# direclty in the WorkGraph. The inputs and outputs of the task is
-# automatically generated based on the input and output port of the AiiDA
-# components.
+# When calling functional tasks in a workgraph, we pass inputs by reference (sockets), not by value.
```
superstar54: Note, we can also pass values directly, e.g.,
```python
multiply(x=1, y=wg.inputs.z).result
```

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,380 @@
...
-# One can also create task from an AiiDA process builder directly.
-#
-
-from aiida.calculations.arithmetic.add import ArithmeticAddCalculation
 
-builder = ArithmeticAddCalculation.get_builder()
-builder.code = code
-builder.x = Int(2)
-builder.y = Int(3)
+@task(outputs=["count", "even_numbers"])
```
superstar54: Because there is a note in the above: `To learn more about sockets, see the Sockets`, I would avoid introducing too much socket concept here.

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,380 @@
...
-wg.run(
-    inputs={"add1": {"x": Int(2), "y": Int(3)}, "multiply1": {"y": Int(4)}},
-)
+# %%
+# We can then call these functional tasks directly in a workgraph as follows:
 
-print("State of WorkGraph:   {}".format(wg.state))
-print("Result of add      : {}".format(wg.tasks.add1.outputs.result.value))
-print("Result of multiply : {}".format(wg.tasks.multiply1.outputs.result.value))
+with WorkGraph("AddMultiplyWithDefinedTasks") as wg:
```
superstar54: I would use `@task.graph` decorator here, so users don't need to know the context manager approach explicitly.
then use the 
```python
wg=func.build_graph()
wg.run()
```

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,380 @@
...
+# We can expose a workgraph as a task by using the ``@task.graph`` decorator.
+# This is useful for task encapsulation and reusability, as well as for defining dynamic (runtime-input-dependent) workflows.
+# Let's see how this works in practice by defining a workflow to sum the even numbers of a number list of random size.
 #
-# The ``Graph Builder`` allow user to create a dynamic workflow based on
-# the input value, as well as nested workflows. Here is an example of
-# nested workflow:
+# .. note::
 #
+#    We'll reuse the ``count_even_numbers`` task defined above
```
superstar54: The `count_even_numbers`, `random_number_generator`, `generate_numbers`, `sum_numbers`, `SumEvenNumbers` tasks introduces additional complex, use `add` and `multiply` only can already cover a simple workflow.

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,380 @@
...
+# How do we control the flow of tasks in ``WorkGraph``?
 
-print("WorkGraph state: ", wg.state)
-print("Result of task add_multiply1: {}".format(outputs2.result.value))
+# %%
+# Control flow
+# ------------
+#
+# Suppose we want to sum a set of numbers conditionally, depending on how many numbers we have.
+# Let's try to define an example workflow using what we've learned so far:
```
superstar54: Better to go to the `@task.graph` approach directly.

superstar54 (changes requested):
Hi @edan-bainglass , thanks for the update! I went through the quick start, and added a few comments. 
Here are some overall suggestions 
- keep the quick start to <= 5 mins read
- use a simple example workflow, so the user can easily understand and run quickly.
- introduce only the essentials needed for the example, try to avoid mentioning advanced concepts

edan-bainglass:
> Hi @edan-bainglass , thanks for the update! I went through the quick start, and added a few comments. Here are some overall suggestions
> 
> * keep the quick start to <= 5 mins read
> * use a simple example workflow, so the user can easily understand and run quickly.
> * introduce only the essentials needed for the example, try to avoid mentioning advanced concepts

Thanks for the review @superstar54. I agree. I'll reduce the quick start section to the simplest @task-based example.

I recommend, however, that we add an introduction section where we cover the following:
- Why use WorkGraph
- The three modes of usage (decorator, context manager, programmatic)
- Maybe more, let's discuss

edan-bainglass:
> > Hi @edan-bainglass , thanks for the update! I went through the quick start, and added a few comments. Here are some overall suggestions
> > 
> > * keep the quick start to <= 5 mins read
> > * use a simple example workflow, so the user can easily understand and run quickly.
> > * introduce only the essentials needed for the example, try to avoid mentioning advanced concepts
> 
> Thanks for the review @superstar54. I agree. I'll reduce the quick start section to the simplest @task-based example.
> 
> I recommend, however, that we add an introduction section where we cover the following:
> 
> * Why use WorkGraph
> * The three modes of usage (decorator, context manager, programmatic)
> * Maybe more, let's discuss

Some of this is already covered in #604. But I still think we need a section where we provide a user with the three modes of usage. Since I've already commented on #604 stating that it should be renamed to Features, perhaps an Overview (or maybe Usage) section could do this.

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,182 @@
...
+# +---------------------------------------------------------+--------------------------------------------------------+
+# | `HowTo <../howto/index.rst>`__                          | Advanced topics and tips, e.g., flow control using     |
+# |                                                         | ``if``, ``for``, ``while``, and ``context``.           |
+# |                                                         |                                                        |
+# +---------------------------------------------------------+--------------------------------------------------------+
+# | `Tutorials <../tutorial/index.rst>`__                   | Real-world examples in computational materials         |
+# |                                                         | science and more.                                      |
+# |                                                         |                                                        |
+# +---------------------------------------------------------+--------------------------------------------------------+
+# | `Migration <migration_from_aiida_core/index.rst>`__     | Migration guide for users interested in transferring   |
```
superstar54:
```suggestion
# | `Migration <../migration_from_aiida_core/index.rst>`__     | Migration guide for users interested in transferring   |
```

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,182 @@
...
+# |                                                         | aiida-core processes (e.g. ``WorkChain``, ``CalcJob``) |
+# |                                                         | to ``WorkGraph``.                                      |
+# |                                                         |                                                        |
```
superstar54, line 187: Also add a link to the GUI.

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,182 @@
...
+# Suppose you want to compute ``(x + y) * z`` as a workflow of two distinct tasks: (i) addition, and (ii) multiplication.
+# Let's see how we can do this using ``WorkGraph``:
 #
+# First, let's define our tasks using the ``@task`` decorator:
 
 from aiida_workgraph import task
 
-# define add task
-@task()
+
```
superstar54, line 60: Delete a empty line, because there are two.
```suggestion
```

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,182 @@
...
-#    to use ``if`` and ``for`` to create the tasks, or repeat a
-#    calculation until it converges.
+# Let's build the workflow with some input and visualize it:
 #
-# The ``Graph Builder`` allow user to create a dynamic workflow based on
-# the input value, as well as nested workflows. Here is an example of
-# nested workflow:
+# .. important::
 #
+#    If you are following along in a Jupyter notebook, omit the ``to_html()`` method call.
```
superstar54: This will not work. One has to replace `wg.to_html()` with `wg`.

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,182 @@
...
 
-from aiida_workgraph import WorkGraph, task
-
-# define add task
-@task()
-def add(x, y):
-    return x + y
+wg = AddMultiply.build_graph(x=2, y=3, z=4)
+wg.to_html()
```
superstar54, line 114:
```suggestion
wg.to_html()
# Comment out the following line to visualize the WorkGraph in the Jupyter notebook.
# wg
```

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,182 @@
...
 
+# %%
+# We can see our two tasks, the assignment of the sum to the multiplication task, and the subsequent assignment of the product to the workflow (graph) result.
+#
+# Let's run the workflow with some inputs and inspect the result:
```
superstar54: There are no inputs here anymore.

docs/gallery/autogen/quick_start.py
```diff
@@ -4,271 +4,182 @@
...
-
-######################################################################
-# Get the result of the tasks:
+# %%
+# .. tip::
 #
+#    If you're running the AiiDA GUI, you can visualize the executed workflow interactively.
+#    Click on the ``PK`` field of the submitted workflow (look for *WorkGraph<AddMultiply>*) to view its details.
+#
+# Let's have a look at the full provenance of our executed workflow:
```
superstar54: Maybe explain what data provenance is and why it's useful, e.g., reproducibility.
```suggestion
# Let's have a look at the full data provenance of our executed workflow:
```

.pre-commit-config.yaml
```diff
@@ -4,6 +4,7 @@ repos:
     hooks:
     -   id: check-yaml
     -   id: end-of-file-fixer
+        exclude: \.svg$
```
superstar54: This is not needed anymore. Don't forget to delete the image too.

superstar54 (approved):
LGTM! I added a few change requests

docs/gallery/autogen/quick_start.py
edan-bainglass: I should maybe say build with some input and run

docs/gallery/autogen/quick_start.py
edan-bainglass: See important note right above it

docs/gallery/autogen/quick_start.py
edan-bainglass: Nevermind. Already built at this point. Okay 👍

docs/gallery/autogen/quick_start.py
superstar54: I see what you mean now. Note, omitting the to_html() call might be confusing, as users could delete the entire line instead of just removing to_html().

edan-bainglass: Okay, I'll be more explicit (verbose) 👍

edan-bainglass:
@superstar54 all should now be resolved. Please have a quick look 🙏

Note: I decided to remove the migration link from "What's next?" for now. It is a very specific thing that maybe is inappropriate to recommend as a next step. We can discuss if it should be included, in which case, we should do it throughout.

superstar54 (approved):
LGTM! Please fix two formatting errors before merging.

image


image

PR #592, opened 2025-07-20
decorator uses dict as socket data

superstar54:
On top of this [PR](https://github.com/scinode/node-graph/pull/70) in the `node-graph`

Because from Python 3.7, the insertion order is guaranteed, so dict is preferred because it makes the code clean and straightforward.

For example, define outputs with an identifier.
```python
@task(
    outputs={
        "sum": {"identifier": "workgraph.Any"},
        "difference": {"identifier": "workgraph.Any"},
    }
)
def add_and_subtract(x, y):
    return {"sum": x + y, "difference": x - y}
```

A list of string is also supported.
```python
@task(outputs=["sum", "difference"])
def add_and_subtract(x, y):
    """This task returns both the sum and difference of two numbers."""
    return {"sum": x + y, "difference": x - y}

```


edan-bainglass:
As mentioned before, regular tasks that define output sockets do not appear to require returning a dictionary with the same keys, but rather allow the user to simply return objects, which are then mapped in order to the defined sockets. I say this because your description explicitly returns a dict.

Also, note that task.graph did not appear to allow the same for some reason.

superstar54:
> I say this because your description explicitly returns a dict.

Good catch~ I have updated the doc: task will map returned values to the  sockets for two cases: dict and tuple

> note that task.graph did not appear to allow the same for some reason.

I opened an [issue](https://github.com/aiidateam/aiida-workgraph/issues/594) on this. Will fix it in a future PR.

edan-bainglass (approved):
LGTM! Thanks 🙏

PR #595, opened 2025-07-21
Map tuple results to the wg outputs

### LINKED ISSUE #588: Inconsistent behavior in task return
Defining output sockets in the task decorator assigns the return arguments by order, avoiding having to manual assign outputs by way of a similarly keyed dictionary. This is great! Unfortunately, it does not work the same for task.graph.

superstar54:
Fix #588, the `graph` task can return a tuple as the result.

```python
@task.graph(outputs=['sum', 'product'])
def add_multiply(x=None, y=None):
    outputs1 = add(x=x, y=y)
    outputs2 = multiply(x=outputs1.result, y=y)
    return outputs1.result, outputs2.result
```

### DIFF
--- a/src/aiida_workgraph/decorator.py
+++ b/src/aiida_workgraph/decorator.py
@@ -71,7 +71,9 @@ def build_task_from_callable(
     raise ValueError(f"The executor {executor} is not supported.")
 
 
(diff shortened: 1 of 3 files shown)


edan-bainglass (commented):
LGTM! Thanks 🙏

edan-bainglass (approved):

PR #597, opened 2025-07-21
use function name as the name of the WorkGraph generated from the task

### LINKED ISSUE #589: Graph name when using `get_graph`
The function name does not transfer. Also, good to provide the user with an API for custom making, e.g., `name` parameter on the decorator

superstar54:
Fix #589


edan-bainglass (approved):
LGTM!

PR #598, opened 2025-07-21
Replace `get_graph` by `build_graph`

### LINKED ISSUE #590: `get_graph` is not so clear
In the new `@task.graph` feature, we can extract the `WorkGraph` instance from the task with `graph_instance = graph_task_instance.get_graph(...)`. However, we need to provide the expected arguments (`...`) to do so. This makes `get_graph` not so clear. `graph_task_instance(...).get_graph()` would be more correct, but then we would be operating on a `NodeSocketNamespace`.

Three solutions I can think of at the moment:
1. Inherit `NodeSocketNamespace` in `GraphSocketNamespace`, which implements `get_graph` to return its parent `WorkGraph`.
2. Keep the method where it is but rename it to `as_graph` or `graph`
3. Provide a class method that can convert the function inputs and outputs to `WorkGraph` inputs/outputs and return the `WorkGraph` instance, thus avoiding having to provide the inputs all together

Comments on solutions:
1 - most reasonable?
2 - only marginally clearer
3 - possibly the ideal case, as it avoids the need for inputs altogether

superstar54:
Fix #590 
`get_graph` gives the impression that the graph already exists. `build_graph` is clear to the user that the graph is built using the given inputs.

### DIFF
--- a/docs/gallery/concept/autogen/graph_builder_concept.py
+++ b/docs/gallery/concept/autogen/graph_builder_concept.py
@@ -56,7 +56,7 @@ def my_workflow(x, y):
 
 
 # A user can now easily create and run the workflow:
-wg = my_workflow.get_graph(x=1, y=2)
+wg = my_workflow.build_graph(x=1, y=2)
 wg.run()
 print("Workflow outputs:", wg.outputs.result)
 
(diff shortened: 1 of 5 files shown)


edan-bainglass (approved):
LGTM! Good to go once tests are fixed.

PR #599, opened 2025-07-21
[Doc] add `Control task execution order`

superstar54:
- merge `zone` and `wait_on` into a new doc: `Control task execution order`.
- add `Zone` context manager
- add `Zone` and `get_current_graph` to the top-level import

Doc preview: https://aiida-workgraph--599.org.readthedocs.build/en/599/howto/autogen/control_task_execution_order.html


docs/gallery/howto/autogen/control_task_execution_order.py
```diff
@@ -0,0 +1,156 @@
...
+
+This tutorial demonstrates two powerful features for managing these dependencies:
+
+* The wait operators: ``>>`` and ``<<``
+* Task grouping with ``Zone``
+
+.. note::
+
+   For dependencies between different WorkGraphs, use the ``monitor`` task.
+   Please refer to the :doc:`monitor <./monitor>` tutorial for more details.
```
GeigerJ2:
```suggestion
   Please refer to the :doc:`monitor <./monitor>` HowTo for more details.
```

docs/gallery/howto/autogen/control_task_execution_order.py
```diff
@@ -0,0 +1,156 @@
...
+# %%
+# Initial setup
+# -------------
+# First, let's set up the environment and adjust the AiiDA log level to ``REPORT`` so we can observe the execution order of the tasks.
+
+from aiida_workgraph.utils.logging import set_aiida_loglevel
+from aiida_workgraph import task, WorkGraph, Zone
+from aiida import load_profile
+
+set_aiida_loglevel("REPORT")
```
GeigerJ2, line 32: Isn't that the default anyway? Is that necessary?

{66473558-00F2-44E6-9BB0-5187287621D7}

docs/gallery/howto/autogen/control_task_execution_order.py
```diff
@@ -0,0 +1,156 @@
...
+# %%
+# Explicit dependencies with wait operators (``>>`` and ``<<``)
+# ==========================================================
+#
+# The simplest way to create an explicit dependency is with the ``>>`` and ``<<``
+# operators. They let you enforce that one task must finish before another begins,
+# even if there's no data link between them.
+#
+# * ``task_A >> task_B`` means **"task B waits for task A"**.
+# * ``task_B << task_A`` means the exact same thing.
```
GeigerJ2: Seems a bit weird to have a list, where the second item only states _means the exact same thing_ ^^ Maybe:

```suggestion
# operators. They let you enforce that one task must finish before another begins,
# even if there's no data link between them.
# To achieve that **"task B waits for task A"**, use the syntax ``task_A >> task_B``, or alternatively, ``task_B << task_A``.
# For the reverse dependency, just swap the tasks.
```

docs/gallery/howto/autogen/control_task_execution_order.py
```diff
@@ -0,0 +1,156 @@
...
+#
+# Example
+# -------
```
GeigerJ2, line 65: Hmm, the TOC seems a bit off:

{5A0DC9CA-D6D1-486C-98AF-3085696F76C5}

Should be:
- Initial setup
- Explicit dependencies with wait operators (<< and <<)
     - Example
- Grouping Dependencies with Zone
     - Example
- Conclusion

No?

docs/gallery/howto/autogen/control_task_execution_order.py
```diff
@@ -0,0 +1,156 @@
...
+wg = wait_graph.build_graph(x=1, y=2)
+wg.run()
+
+# %%
+# By checking the ``REPORT`` logs from AiiDA, you will see that both ``multiply``
+# tasks complete before the ``add`` task begins, just as we specified.
+
+
+# %%
+# Grouping Dependencies with ``Zone``
```
GeigerJ2:
```suggestion
# Grouping dependencies with ``Zone``
```
Think we agreed on making only the first word capitalized?

docs/gallery/howto/autogen/control_task_execution_order.py
```diff
@@ -0,0 +1,156 @@
...
+# ===================================
+#
+# For more complex scenarios, you can group a set of tasks into a **Zone**.
+# A ``Zone`` acts as a single unit for dependency management, governed by two rules:
+#
+# 1. **Entry Condition**: A ``Zone`` (and all tasks within it) will only start
+#    after *all* tasks with links pointing *into* the ``Zone`` are finished.
+# 2. **Exit Condition**: Any task that needs an output from *any* task inside the
+#    ``Zone`` must wait for the entire `Zone` to complete.
+#
```
GeigerJ2, line 113:
```suggestion
#
# To add tasks to a `Zone`, we use the context manager syntax, in analogy to the WorkGraph.
# In the example below, all tasks that are contained in the indented context manager code block belong to the active `Zone` within.
```

What about also introducing task syntax here, e.g., `@task.zone`? Similar to how we did now for the WG, where the context manager is implicit in `@task.graph`? Or is this already possible?

docs/gallery/howto/autogen/control_task_execution_order.py
```diff
@@ -0,0 +1,156 @@
...
+# * Use the wait operators (``>>``, ``<<``) for simple, direct dependencies
+#   between individual tasks.
+# * Use a ``Zone`` to group tasks that should be treated as a single execution
+#   block, simplifying complex dependency chains.
+#
+# These tools give you fine-grained control over your workflows, ensuring your
+# calculations run in the correct sequence.
+
+# Reset the log level to avoid verbose output in other examples
+set_aiida_loglevel("ERROR")
```
GeigerJ2, line 162:
```suggestion
```

GeigerJ2 (approved):
LGTM, just some minor comments, but approving already!

docs/gallery/howto/autogen/control_task_execution_order.py
GeigerJ2: Yes, I just tested this, and if the `logging.aiida_loglevel` is still at the AiiDA default, the output is the same. So these parts of the code can be removed.

superstar54: We set the loglevel to `ERROR` in the configuration ([here](https://github.com/aiidateam/aiida-workgraph/blob/4cd477cdf725c7fa5219490234e54385be0a3cdb/.readthedocs.yaml#L19)) of the docs to avoid printing out too many logs when using `wg.run()`. If some doc needs the logs info, then we can switch it to `REPORT` only for this doc.

docs/gallery/howto/autogen/control_task_execution_order.py
superstar54: I refactored the text a little bit, and keep the list. :laughing:

docs/gallery/howto/autogen/control_task_execution_order.py
superstar54: This is not possible yet, but it is indeed something we could consider in the future.

docs/gallery/howto/autogen/control_task_execution_order.py
GeigerJ2: Ah, OK, I see. Thanks for the explanation!

PR #602, opened 2025-07-22
Fix `required metadata` when constructing ports

t-reents:
The `required` metadata wasn't correctly passed, causing jobs to except if non-required inputs or outputs were not provided. 

I was working on this branch for quite some time, so this change seems to work.


superstar54 (approved):
Thanks for the fix! LGTM!

PR #605, opened 2025-07-23
Support aiida-core 2.7

### LINKED ISSUE #481: Test for aiida-core 2.7.0
Test for aiida-core 2.7.0, and consider which aiida-core versions to be supported.

superstar54:
I ran all the tests locally, and only one test failed. The test failed because the api to interact with the process is changed in this [PR](https://github.com/aiidateam/aiida-core/pull/6902) and in aiida-core >=2.7.

In this PR, I have updated the API, and set the `aiida-core` version to `~=2.6`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
     "ase",
     "node-graph==0.2.16",
     "node-graph-widget>=0.0.5",
-    "aiida-core>=2.3,<=2.6.3",
+    "aiida-core>=2.3,~=2.6",
     "cloudpickle",
     "aiida-shell~=0.8",
     "aiida-pythonjob>=0.2.3",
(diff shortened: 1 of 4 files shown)


edan-bainglass (approved):
LGTM! Glad it was 99% compatible already :)

t-reents (approved):
Looks good (I already tested my production runs with `aiida-core 2.7` yesterday, and everything seems fine)

pyproject.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
     "ase",
     "node-graph==0.2.16",
     "node-graph-widget>=0.0.5",
-    "aiida-core>=2.3,<=2.6.3",
+    "aiida-core>=2.3,~=2.6",
```
GeigerJ2, line 33: These requirements don't make sense, though, no? `~=2.6` means `>=2.6.0` plus `<3.0.0` (which will probably never happen 🥲), hence, the `>=2.3` is redundant. Or am I getting something wrong?

superstar54: Good catch! I was intending to change it to `aiida-core~=2.6`.

PR #607, opened 2025-07-23
Clean unused doc files

superstar54:
During the refactor of documentation, there are many unused files left.


edan-bainglass (approved):
Solid clean up 👍

PR #610, opened 2025-07-26
Add basic tests for cli

superstar54:
Although the CLI is not yet useful, we should add some basic tests to ensure it does not break.
See this : https://github.com/conda-forge/aiida-workgraph-feedstock/pull/15#issuecomment-3121775586

This RP:
- fixes the CLI when using aiida-core 2.7, same as this PR: https://github.com/aiidateam/aiida-workgraph/pull/605.
- add basic tests

For an unknown reason, the `test_cli` affects the report of the Process. The `get_workchain_report` gives 'No log messages recorded for this entry'. A temporary solution is to capture the output stream directly and check the output.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,7 +37,7 @@ jobs:
         strategy:
             matrix:
                 python-version: ['3.9', '3.11', '3.12']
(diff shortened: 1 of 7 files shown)


edan-bainglass (approved):
LGTM!

PR #613, opened 2025-07-29
Fix top level outputs link

superstar54:
If the top-level outputs is used in a link, we use the built-in `_outputs` socket to represent it.

### DIFF
--- a/src/aiida_workgraph/workgraph.py
+++ b/src/aiida_workgraph/workgraph.py
@@ -593,6 +593,10 @@ class WorkGraph(node_graph.NodeGraph):
         """Add a link between two tasks."""
         if isinstance(source, Task):
             source = source.outputs["_outputs"]
+        elif source._parent is None:
+            # if the source is the top-level outputs,
+            # we use the built-in "_outputs" socket to represent it
+            source = source._outputs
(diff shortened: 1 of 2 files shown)


edan-bainglass:
> If the top-level outputs is used in a link, we use the built-in `_outputs` socket to represent it.

Can you explain this a bit more, perhaps with a use-case/example?

superstar54:
Yes, here is an example:
```python
@task
def add_multily(x, y):

    return {"sum": x + y, "product": x * y}

@task
def total(**data):
    """Calculate the total from a dictionary of values."""
    return sum(data.values())

with active_graph(WorkGraph()) as wg:
    outputs = add_multily(x=1, y=2)
    # here we pass the outputs directly into the next task, instead of outputs.sum, or outputs.product.
    outputs = total(data=outputs)
wg
```
Here is the representation in GUI
image

The built-in socket `_outputs` is used to pass the whole outputs (as a dict) into the next task.

edan-bainglass:
@superstar54 it's not immediately clear why this is (was? 😅) required. What happened in your above example prior to this change?

superstar54:
> What happened in your above example prior to this change?

It will raise an error: `"outputs" is not in the TaskSocketNamespace`. 
All the built-in sockets are start with `_`. For the moment, there are:
- `_outputs` represent the whole `outputs`
- `_wait`

PR #614, opened 2025-07-29
Auto detect and link graph inputs for `graph` tasks

### LINKED ISSUE #603: Auto-detect graph-level inputs when using `.build_graph`
Take the following decorator-based graph definition:

```python
@task.graph
def AddMultiply(x, y, z):
    the_sum = add(x=x, y=y).result
    return multiply(x=the_sum, y=z).result
```

If we do `AddMultiply.build_graph(1, 2, 3)` and visualize it, we get

Image

Graph-level outputs are correctly registered, but not the graph-level inputs. It would be good to have this auto detected and added. However, as mentioned by @superstar54, it is more work (possibly requiring `inspect` and `ast`) to detect the links of said inputs with the encapsulated tasks.

Something to think about.

superstar54:
Resolves #603.

### Pros

* Clear visualization of graph inputs, similar to the context manager approach.
* When the WorkGraph runs, graph-level inputs are serialized and passed to tasks, so the data flow within the AiiDA process can be tracked. Otherwise, tasks would serialize their raw inputs, creating multiple distinct AiiDA data nodes for the same data.
* Serialization requires users to declare the input structure, which makes graphs more robust and less error prone.

### Cons

* The serialization requirement forces users to declare the input structure, which raises the learning curve for new users. They also need to understand the concept of namespaces.

## Implementation

There are many ways to track inputs. One option is static analysis with Python’s `ast` module, as mentioned in #603. However, static analysis cannot see branches and values that only materialize at runtime.

In this PR, we record the `id` of input values and build `value` to `id` mappings for the graph-level inputs and for every task’s inputs. Using these mappings, we can detect which tasks each input value flows into. Since this happens at runtime, it captures complex control flow.

Example:

```python
@task
def add(x, y):
    return x + y

@task
def multiply(x, y):
    return x * y

@task.graph
def add_multiply(a, b, c):
    sum = add(x=a, y=b).result
    if a > 0:
        a = add(x=a, y=b).result
        product = multiply(a, y=c)
    else:
        product = multiply(a, y=c)
    return product

wg = add_multiply.build_graph(a=1, b=2, c=3)
wg
```

If `a = -1`, the branch and data flow change accordingly.

## Limitations of the current approach

In Python:

```python
x = 1
y = x
add(1, y=y)
```

Here `x`, `y`, and `1` can share the same identity, so you cannot distinguish variable names from the literal value using identity alone. Using `ast` helps recover variable names from source, but covering all real-world cases is difficult, and correctness is hard to guarantee.

## Alternative approach

* Define `inputs` in the decorator.
* Inside the function, use `wg.inputs` explicitly.

This is opt in. If users do not reference `wg.inputs`, this mode is not activated and `wg.inputs` is cleared.

```python
@task.graph(inputs={"nested": {"identifier": "workgraph.namespace",
                               "sockets": {"a": {"identifier": "any"},
                                           "b": {"identifier": "any"}}}})
def AddMultiply(x, y, nested: dict):
    wg = get_current_graph()
    x = add(x=wg.inputs.x, y=wg.inputs.y).result
    y = add(x=wg.inputs.nested["a"], y=wg.inputs.nested["b"]).result
    return multiply(x=x, y=y).result

wg = AddMultiply.build_graph(
    x=1,
    y=2,
    nested={"a": 1, "b": 3},
)
```

### DIFF
--- a/src/aiida_workgraph/decorator.py
+++ b/src/aiida_workgraph/decorator.py
@@ -99,7 +99,7 @@ def _assign_wg_outputs(
 
 def _run_func_with_wg(
     func: Callable,
-    graph_task_output_names: List[str],
+    TaskCls: callable,
     args: tuple,
     kwargs: dict,
     var_kwargs: Optional[dict] = None,
@@ -110,9 +110,46 @@ def _run_func_with_wg(
     """
     merged = {**kwargs, **(var_kwargs or {})}
     with WorkGraph(func.__name__) as wg:
+        wg.graph_inputs.inputs = TaskCls.InputCollectionClass._from_dict(
+            TaskCls._ndata.get("inputs", {}),
+            node=wg.graph_inputs,
+            pool=TaskCls.SocketPool,
+            graph=wg,
+        )
+        graph_task_output_names = [
+            name
+            for name, socket in TaskCls._ndata.get("outputs", {})
+            .get("sockets", {})
+            .items()
+            if not socket.get("metadata", {}).get("builtin_socket", False)
+        ]
+        wg.inputs = prepare_function_inputs(func, *args, **merged)
         raw = func(*args, **merged)
         _assign_wg_outputs(raw, wg, graph_task_output_names)
         return wg
+
+
+def prepare_function_inputs(func, *call_args, **call_kwargs):
+    """Prepare the inputs for the function call.
+    This function extracts the arguments from the function signature and
+    assigns them to the inputs dictionary."""
+    inputs = dict(call_kwargs or {})
+    if func is not None:
+        arguments = list(call_args)
+        orginal_func = func._func if hasattr(func, "_func") else func
+        for name, parameter in inspect.signature(orginal_func).parameters.items():
+            if parameter.kind in [
+                parameter.POSITIONAL_ONLY,
+                parameter.POSITIONAL_OR_KEYWORD,
+            ]:
+                try:
+                    inputs[name] = arguments.pop(0)
+                except IndexError:
+                    pass
+            elif parameter.kind is parameter.VAR_POSITIONAL:
+                # not supported
+                raise ValueError("VAR_POSITIONAL is not supported.")
+    return inputs
 
 
 def _make_wrapper(TaskCls, func=None):
@@ -133,23 +170,7 @@ def _make_wrapper(TaskCls, func=None):
         if active_zone:
             active_zone.children.add(task)
 
-        inputs = dict(call_kwargs or {})
-        if func is not None:
-            arguments = list(call_args)
-            orginal_func = func._func if hasattr(func, "_func") else func
-
-            for name, parameter in inspect.signature(orginal_func).parameters.items():
-                if parameter.kind in [
-                    parameter.POSITIONAL_ONLY,
-                    parameter.POSITIONAL_OR_KEYWORD,
-                ]:
-                    try:
-                        inputs[name] = arguments.pop(0)
-                    except IndexError:
-                        pass
-                elif parameter.kind is parameter.VAR_POSITIONAL:
-                    # not supported
-                    raise ValueError("VAR_POSITIONAL is not supported.")
+        inputs = prepare_function_inputs(func, *call_args, **call_kwargs)
         task.set(inputs)
         return task.outputs
 
@@ -293,15 +314,8 @@ class TaskDecoratorCollection:
 
             def build_graph(*args, **kwargs):
                 """This function is used to get the graph from the wrapped function."""
-                graph_task_output_names = [
-                    name
-                    for name, socket in TaskCls._ndata.get("outputs", {})
-                    .get("sockets", {})
-                    .items()
-                    if not socket.get("metadata", {}).get("builtin_socket", False)
-                ]
 
-                return _run_func_with_wg(func, graph_task_output_names, args, kwargs)
+                return _run_func_with_wg(func, TaskCls, args, kwargs)
 
             wrapped_func.build_graph = build_graph
 
--- a/src/aiida_workgraph/decorator.py
+++ b/src/aiida_workgraph/decorator.py
@@ -13,6 +13,7 @@ from aiida_workgraph.tasks.factory import (
     PyFunctionTaskFactory,
 )
 from node_graph.utils import socket_value_id_mapping
+from node_graph.socket import NodeSocket
 
 
 def build_task(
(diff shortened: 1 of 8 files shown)


edan-bainglass (approved):
LGTM! Once the tests pass... 😅

edan-bainglass:
@superstar54 hmm, there are cases where this doesn't work as expected. I'll list them here:

### Input order

```python
@task.graph
def WhileLoop(n, m):
    if m >= n:
        return m
    m = add(x=m, y=1).result
    return WhileLoop(n=n, m=m)


WhileLoop.build_graph(n=4, m=0)
```
image

Should the inputs follow the order of signature definition?

### Incorrect mapping

```python


@task
def add(x, y):
    return x + y


@task
def multiply(x, y):
    return x * y


@task.graph
def AddMultiply(x, y, z):
    the_sum = add(x=x, y=y).result
    return multiply(x=the_sum, y=z).result


@task
def generate_random_number(minimum, maximum):
    import random

    return random.randint(minimum, maximum)


@task.graph
def AddMultiplyComposed(minimum, maximum, x, y):
    random_number = generate_random_number(minimum=minimum, maximum=maximum).result
    return AddMultiply(x=x, y=y, z=random_number).result


wg = AddMultiplyComposed.build_graph(minimum=1, maximum=10, x=1, y=2)
```
image

Somehow the minimum argument got mixed with the x argument.

### Monitor task

```python
import datetime


@task.monitor
def monitor_time(time):


@task
def add(x, y):
    return x + y


@task.graph
def TimeMonitor(time, x, y):
    monitor_time(time) >> (the_sum := add(x, y).result)  # wait THEN (>>) add
    return the_sum


TimeMonitor.build_graph(
    time=datetime.datetime.now() + datetime.timedelta(seconds=5),
    x=1,
    y=2,
)
```
image

No `graph_inputs` 🥲

edan-bainglass:
> ### Incorrect mapping
> ```python
> from aiida_workgraph import task
> 
> 
> @task
> def add(x, y):
>     return x + y
> 
> 
> @task
> def multiply(x, y):
>     return x * y
> 
> 
> @task.graph
> def AddMultiply(x, y, z):
>     the_sum = add(x=x, y=y).result
>     return multiply(x=the_sum, y=z).result
> 
> 
> @task
> def generate_random_number(minimum, maximum):
>     import random
> 
>     return random.randint(minimum, maximum)
> 
> 
> @task.graph
> def AddMultiplyComposed(minimum, maximum, x, y):
>     random_number = generate_random_number(minimum=minimum, maximum=maximum).result
>     return AddMultiply(x=x, y=y, z=random_number).result
> 
> 
> wg = AddMultiplyComposed.build_graph(minimum=1, maximum=10, x=1, y=2)
> ```
> 
> image
> Somehow the minimum argument got mixed with the x argument.

I see the issue. You use `id(value)` as the unique identifier `_value_id`. But this is not unique. In the above case, they are both `1`, which explains the mix up. Would the property name be sufficient as an identifier in the mapping? Or is this not unique?

edan-bainglass:
> I see the issue. You use `id(value)` as the unique identifier `_value_id`. But this is not unique. In the above case, they are both `1`, which explains the mix up. Would the property name be sufficient as an identifier in the mapping? Or is this not unique?

Using the `name` as the mapping key resolves the mix-up described above, as well as restores correct inputs for the monitor task. However, it also breaks the while loop case.

image

I think maybe the link resolution function isn't taking recursion into account. I tried implementing this but got a bit lost in your repo 😅

edan-bainglass:
Okay, never mind. Maybe `id(value)` is right. If I understand this correctly, you are trying to map the input sockets of the graph to the input sockets of any task that takes the graph input socket by value. For example:

```python
@task.graph
def AddMultiply(x, y, z):  # graph input socket z...
    the_sum = add(x=x, y=y).result
    return multiply(x=the_sum, y=z).result  # ...is assigned as the VALUE of the multiply input socket y
```

A graph may have several tasks that receive `z` as an input VALUE and use the mapping to resolve this.

### The real issue

In the above example, if you provide `x=1` and `y=1`, you get `x` links to both `add.x` AND `add.y`, because it cannot differentiate the values. It sees that `x` is 1 and that 1 goes into `x` AND `y`, so it assumes the links.

The present solution considers `id(value)` to accommodate runtime logic (dynamic workflows), I assume. This is indeed a requirement! But the current solution must differentiate values coming from DIFFERENT inputs.

src/aiida_workgraph/decorator.py
```diff
@@ -98,16 +99,82 @@ def _assign_wg_outputs(
...
+        for s in potential_sources:
+            if s._name == source_variable_name:
+                return s
+        # If we couldn't find a match, it's likely due to variable aliasing
+        # (e.g., `graph_input(x=1); y=x; task(a=y)`). The link from `x` to `a`
+        # cannot be inferred automatically.
+        warnings.warn(
+            f"Could not automatically resolve link for input '{task_socket._name}' of task '{task_name}'. "
+            f"This can happen if the input variable was renamed (e.g., 'y=x; task(a=y)'). "
+            f"The calculation will proceed correctly, but the graph visualization may be incomplete."
```
edan-bainglass, line 143: I think this is fair. It is an edge case that represents possibly not the best practice. In the absence of a better solution, we just document that variable aliasing can lead to wrong graph visualization.

edan-bainglass (commented):
Just one comment for now. Need to test.

edan-bainglass:
> ### Pros
> * Clear visualization of graph inputs, similar to the context manager approach.
> * When the WorkGraph runs, graph-level inputs are serialized and passed to tasks, so the data flow within the AiiDA process can be tracked. Otherwise, tasks would serialize their raw inputs, creating multiple distinct AiiDA data nodes for the same data.
> * Serialization requires users to declare the input structure, which makes graphs more robust and less error prone.

Solid pros!

> ### Cons
> * The serialization requirement forces users to declare the input structure, which raises the learning curve for new users. They also need to understand the concept of namespaces.

As you say, this is opt-in. Do more to get more. However, if they do less, we need to make sure an incorrect state is not produced. I believe you use a warning when aliasing is detected, in which case, what would the graph look like?

edan-bainglass:
It is still incorrect, even without aliasing. Consider the followin:

```python
import os
import asyncio
from aiida_workgraph.collection import group


@task.awaitable
async def sleep_create_file(filepath, content):
    await asyncio.sleep(5)
    with open(filepath, "w") as f:
        f.write(content)


@task.monitor
def monitor_file(filepath):
    return os.path.exists(filepath)


@task
def discard_file(filepath):
    if os.path.exists(filepath):
        os.remove(filepath)
    else:
        raise FileNotFoundError(f"File {filepath} does not exist.")


@task
def multiply(x, y):
    return x * y


@task.graph
def FileMonitor(x, y, z):
    # Asynchronously create a file after 5 seconds
    sleep_create_file(
        filepath="/tmp/monitor_test.txt",
        content="This is a test file",
    )

    # While above is running, monitor for the file
    # Once done (`>>` means wait on left operant), add two numbers and discard the file
    # Finally, multiply the sum by a factor
    (
        monitor_file(
            "/tmp/monitor_test.txt",
            interval=1,
            timeout=10,
        )
        >> group(
            discard_file("/tmp/monitor_test.txt"),
            the_sum := add(x, y).result,
        )
        >> (the_product := multiply(the_sum, z).result)
    )

    return the_product


wg = FileMonitor.build_graph(x=1, y=2, z=3)
wg
```

which produces

image

Again, `x=1`, but also the monitoring interval is `1`, so `x` is incorrectly linked to both 😬 Everything else appears to be correct though, which is good.

superstar54:
Thanks @edan-bainglass  for reviewing this PR!

Superseded by #627

PR #616, opened 2025-07-29
Fix graph-level sockets

### LINKED ISSUE #615: `WorkGraph` restart feature is broken
Running/submitting a restarted workgraph no longer works due to an issue with graph-level sockets.

superstar54:
- The `node_class` of graph-level tasks should be itself, so it can be reloaded correctly from the database.
- In the engine, initialize the results of the graph-level sockets first, so other tasks can update them.


edan-bainglass (approved):
Looks good 👍 Merge after test is fixed?

PR #622, opened 2025-07-30
Switch a few sections already to decorator-paradigm

edan-bainglass:
Already switching some sections where simple/clear to do so.

### DIFF
--- a/docs/gallery/howto/autogen/async.py
+++ b/docs/gallery/howto/autogen/async.py
@@ -9,18 +9,16 @@ Run ``async`` functions as tasks
 #
 # The ``awaitable`` decorator allows for the integration of ``asyncio`` within tasks, letting users control asynchronous functions.
 
-# %%
-# We'll temporarily set the AiiDA log level to ``REPORT``, so that we can inspect the execution order of the workgraph.
-
+# sphinx_gallery_start_ignore
 from aiida_workgraph.utils.logging import set_aiida_loglevel
 
 set_aiida_loglevel("REPORT")
+# sphinx_gallery_end_ignore
 
-# %%
 import asyncio
 
 from aiida import load_profile
-from aiida_workgraph import WorkGraph, task
+from aiida_workgraph import task
 
 load_profile()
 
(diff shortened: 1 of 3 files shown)


docs/gallery/howto/autogen/async.py
```diff
@@ -29,28 +27,39 @@
...
+    await asyncio.sleep(4)
     return x + y
 
 
-with WorkGraph("AwaitableGraph") as wg:
-    wg.inputs = dict.fromkeys(["x", "y"])
-    awaitable_task(x=wg.inputs.x, y=wg.inputs.y)
-    wg.inputs.x + wg.inputs.y
+@task.graph
+def AwaitableSum(x, y):
```
superstar54, line 47: For the function name, it's better to use snake_case.

docs/gallery/howto/autogen/async.py
```diff
@@ -9,18 +9,16 @@
 #
 # The ``awaitable`` decorator allows for the integration of ``asyncio`` within tasks, letting users control asynchronous functions.
 
-# %%
-# We'll temporarily set the AiiDA log level to ``REPORT``, so that we can inspect the execution order of the workgraph.
-
+# sphinx_gallery_start_ignore
```
superstar54, line 12: This is nice!

docs/gallery/howto/autogen/monitor.py
```diff
@@ -48,33 +49,30 @@
...
+
+
+@task
+def add(x, y):
+    return x + y
+
 
+@task.graph
+def TimeMonitor(time, x, y):
+    monitor_time(time) >> (the_sum := add(x, y).result)  # wait THEN (>>) add
```
superstar54, line 63: Many people may not know this `:=`, but it's fine to introduce it here.

superstar54 (changes requested):

docs/gallery/howto/autogen/async.py
edan-bainglass: Well, not exactly. Need to think about these as workflows. Two options: (i) we follow the WorkChain convention and name them as nouns, e.g., AddMultiply, in which case we use camel case; (ii) we follow the task convention and name them as verbs, e.g., add_and_multiply, in which case we use snake case. I'm good either way. Which do you prefer?

Note that naming functions using snake case is a python convention when functions are intended to be used as such, functionally. But python is flexible. For example, functions are used as components in Solara, in which case, they are not an action, but a thing! Hence, camel case 🐫

docs/gallery/howto/autogen/monitor.py
edan-bainglass: I'm fine rewriting it without it.

docs/gallery/howto/autogen/async.py
superstar54: It's fine for me to name the `graph` task using camel case, and all other function tasks using snake case. Good that we keep this convention through the docs, and also recommend the user to do this. But this can be done in a future PR.

superstar54 (approved):
LGTM!

PR #623, opened 2025-07-30
Add built-in monitor tasks via decotrated functions

edan-bainglass:
This PR replaces the class-based monitor tasks with decorated functional tasks.


edan-bainglass:
@superstar54 not sure if there is use anymore for the classes in task/monitors.py. Can the entry points just point at the decorated functions?

superstar54:
You are right, there is no need to define the Task class manually anymore. The entry points can point to the decorated functions directly. Could you update this in this PR?

PR #627, opened 2025-08-12
Use `TaggedValue` to track the graph inputs

### LINKED ISSUE #603: Auto-detect graph-level inputs when using `.build_graph`

superstar54:
Resolves #603. supercedes #614 

### Pros

* Clear visualization of graph inputs, similar to the context manager approach.
* When the WorkGraph runs, graph-level inputs are serialized and passed to tasks, so the data flow within the AiiDA process can be tracked. Otherwise, tasks would serialize their raw inputs, creating multiple distinct AiiDA data nodes for the same data.
* Serialization requires users to declare the input structure, which makes graphs more robust and less error prone.

### Cons

* The serialization requirement forces users to declare the input structure, which raises the learning curve for new users. They also need to understand the concept of namespaces.

## Implementation

In this PR, we use the `TaggedValue` class, which transparently behaves like the wrapped value, as well as attach a `_socket` attribute referencing its originating graph input socket.
When a `TaggedValue` is passed to a task's socket, a link will be made instead of assigning the value. Since this happens at runtime, it captures complex control flow.
This should be robust and cover all possible cases.

## 

Example:

```python
@task
def add(x, y):
    return x + y

@task
def multiply(x, y):
    return x * y

@task.graph
def add_multiply(a, b, c):
    sum = add(x=a, y=b).result
    if a > 0:
        a = add(x=a, y=b).result
        product = multiply(a, y=c)
    else:
        product = multiply(a, y=c)
    return product

wg = add_multiply.build_graph(a=1, b=2, c=3)
wg
```

## Possible limitation
A potential limitation is whether the `TaggedValue` class is general enough to wrap all types of Python objects. The implementation uses the [wrapt](https://github.com/GrahamDumpleton/wrapt) package, which is a widely-adopted library (used by over 601,000 packages on GitHub). This widespread usage suggests that it is robust enough for this application.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,11 +4,11 @@ build-backend = "flit_core.buildapi"
 
 [project]
 name = "aiida-workgraph"
-dynamic = ["version"] # read from aiida_workgraph/__init__.py
+dynamic = ["version"]  # read from aiida_workgraph/__init__.py
 description = "Design flexible node-based workflow for AiiDA calculation."
-authors = [{ name = "Xing Wang", email = "xingwang1991@gmail.com" }]
+authors = [{name = "Xing Wang", email = "xingwang1991@gmail.com"}]
 readme = "README.md"
-license = { file = "LICENSE" }
+license = {file = "LICENSE"}
 classifiers = [
     "Development Status :: 1 - Planning",
     "Framework :: AiiDA",
@@ -28,7 +28,7 @@ dependencies = [
     "numpy~=1.21",
     "scipy",
     "ase",
-    "node-graph==0.2.19",
+    "node-graph @ git+https://github.com/scinode/node-graph.git",
     "node-graph-widget>=0.0.5",
     "aiida-core~=2.6",
     "cloudpickle",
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ dependencies = [
     "numpy~=1.21",
     "scipy",
     "ase",
-    "node-graph==0.2.19",
+    "node-graph @ git+https://github.com/superstar54/node-graph.git@use_wrapt_proxy",
```
superstar54: This is only used to test. will change it back when node-graph publishes a new version.

edan-bainglass (approved):
LGTM! Excellent work @superstar54 👏

PR #632, opened 2025-08-14
Update PythonJob using new spec API

superstar54:
We have a new `spec` api, and we also used it in the `aiida-pythonjob` in this [PR](https://github.com/aiidateam/aiida-pythonjob/pull/35). This PR updates the related code for the new API.

### DIFF
--- a/src/aiida_workgraph/tasks/factory/pythonjob/tasks.py
+++ b/src/aiida_workgraph/tasks/factory/pythonjob/tasks.py
@@ -9,6 +9,10 @@ from aiida_workgraph.utils import create_and_pause_process
 from node_graph.executor import NodeExecutor
 from aiida.engine import run_get_node
 from aiida_pythonjob import pyfunction
+from aiida_pythonjob.ports_adapter import (
+    inputs_sockets_to_ports,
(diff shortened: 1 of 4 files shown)


src/aiida_workgraph/tasks/factory/pythonjob/tasks.py
edan-bainglass, line 1: I raised a question about `ports_adapter` in the referenced PR. Can you also say a bit here why/how you're using its functions?

edan-bainglass (commented):
Only comments for now 🙂

src/aiida_workgraph/tasks/factory/pythonjob/tasks.py
superstar54: Yes, these functions convert the socket data to the port data, which is used by the Python job. `Socket` and `Port` are very similar schemas, but `Socket` has more metadata (e.g., links).

I have added a more detailed explanation in the referenced PR.

edan-bainglass: Thanks 🙂

edan-bainglass (approved):
LGTM!

PR #634, opened 2025-08-15
Add AiiDA node for `datetime.datetime`

### LINKED ISSUE #620: `datetime` object now serializable?
```python
@task.monitor
def monitor_time(time):


@task
def add(x, y):
    return x + y


@task.graph
def TimeMonitor(time, x, y):
    monitor_time(time) >> add(x, y)  # wait THEN (>>) add


wg = TimeMonitor.build_graph(
    time=datetime.datetime.now() + datetime.timedelta(seconds=5),
    x=1,
    y=2,
)
wg.run()
```
This works now, which is great! However, it didn't used to. Not too long ago, I had to convert the datetime object to ISO format, then back into a `datetime` object in the monitor task.

Opening this as a reminder to check what changed.

### LINKED ISSUE #626: Create a AiiDA data node for `datetime.datetime` for the monitor

superstar54:
Add a `DateTimeData` AiiDA node for `datetime.datetime`.

### DIFF
--- a/docs/gallery/howto/autogen/monitor.py
+++ b/docs/gallery/howto/autogen/monitor.py
@@ -49,9 +49,8 @@ import datetime
 
 
 @task.monitor
-def monitor_time(time: str):
-    time = datetime.datetime.fromisoformat(time)
-    return datetime.datetime.now() > time
+def monitor_time(time: datetime.datetime):
+    return datetime.datetime.now() > time.value
 
 
 @task
@@ -66,7 +65,7 @@ def TimeMonitor(time, x, y):
 
 
 wg = TimeMonitor.build_graph(
-    time=(datetime.datetime.now() + datetime.timedelta(seconds=5)).isoformat(),
+    time=datetime.datetime.now() + datetime.timedelta(seconds=5),
     x=1,
     y=2,
 )
--- a/docs/source/overview.rst
+++ b/docs/source/overview.rst
@@ -153,7 +153,7 @@ Advanced control and modularity
         def time_monitor(time):
             """This task waits until a specified time has passed."""
             import datetime
-            return datetime.datetime.now() > datetime.datetime.fromisoformat(time.value)
+            return datetime.datetime.now() > time.value
 
 Intuitive user experience
 -------------------------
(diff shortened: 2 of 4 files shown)


docs/gallery/howto/autogen/monitor.py
```diff
@@ -49,9 +49,8 @@
 
 
 @task.monitor
-def monitor_time(time: str):
-    time = datetime.datetime.fromisoformat(time)
-    return datetime.datetime.now() > time
+def monitor_time(time: datetime.datetime):
+    return datetime.datetime.now() > time.value
```
edan-bainglass, line 53: It is inaccurate and misleading to annotate `time` as `datetime.datetime` and then do `time.value`. For those familiar with `datetime`'s API, `value` wouldn't make much sense. Of course, it is there because it is really a `DatatimeData`.

edan-bainglass: I'm still perplexed by the fact that at one point in time, my docs were compiling and running without having to serialize the `time` argument. Unfortunately, I cannot reproduce 🥲

edan-bainglass: How does it work with strings, for example? We serialize to AiiDA's `Str`, no? But we don't have to do `string.value`?

edan-bainglass (commented):
A few questions 🙂

docs/gallery/howto/autogen/monitor.py
superstar54: You caught the devil. 
- normal task (underlying PyFunction), it automatically serializes and deserializes the AiiDA node, so the user only needs to deal with the raw Python data. The task also has a corresponding AiiDA process node.
- monitor task, and the awaitable task (async function), are currently handled by the WorkGraph engine, and do have their own AiiDA process node, and also do not handle serialization/deserialization. That's why one has to use `time.value`.

In the future, we can add serialization/deserialization to the monitor task, too.

> We serialize to AiiDA's Str, no? But we don't have to do string.value?

I think for the basic type, it just works like the raw data.

edan-bainglass: Okay 😈

edan-bainglass (approved):
LGTM!

PR #635, opened 2025-08-15
Improve graph level inputs/outputs api using `spec`

### LINKED ISSUE #584: Improve graph-level inputs api
At the moment, you can do `wg.inputs = {"a": None, "b": None}`, which you can streamline with `wg.inputs = dict.fromkeys(["a", "b"])`.

I propose we replace the api with the following:

1. `wg.inputs = ["a", "b"]` to be auto-detected internally and converted into `dict.fromkeys`
2. `wg.inputs = {"a": {...}, "b": {...}}`, where `...` is the usual metadata you'd provide to `spec.input` in `aiida-core` (see for example [here](https://aiida.readthedocs.io/projects/aiida-core/en/stable/topics/processes/usage.html#validation-and-defaults))

@superstar54, @GeigerJ2

superstar54:
See the description of this node-graph [PR](https://github.com/scinode/node-graph/pull/79).

Example usage can be found in this [section](https://aiida-workgraph--635.org.readthedocs.build/en/635/advanced/autogen/context_manager.html)

### DIFF
--- a/docs/gallery/advanced/autogen/context_manager.py
+++ b/docs/gallery/advanced/autogen/context_manager.py
@@ -121,14 +121,6 @@ with WorkGraph(
         sums=spec.namespace(first=any, second=any, third=any), product=any
     ),
 ) as wg:
-    wg.inputs = {
-        "add": {
-            "first": dict.fromkeys(["x", "y"]),
-            "second": dict.fromkeys(["x", "y"]),
-        },
-        "multiply": dict.fromkeys(["factor"]),
-    }
-
     first_sum = add(
         x=wg.inputs.add.first.x,
         y=wg.inputs.add.first.y,
--- a/docs/gallery/advanced/autogen/context_manager.py
+++ b/docs/gallery/advanced/autogen/context_manager.py
@@ -49,11 +49,11 @@ def AddMultiply(x, y, z):
 # %%
 # Now, let's see how we can achieve the same result using the context manager paradigm.
 
-from aiida_workgraph import WorkGraph
+from aiida_workgraph import WorkGraph, spec
 
-with WorkGraph("AddMultiplyContextManager") as wg:
-    wg.inputs = dict.fromkeys(["x", "y", "z"])
-
+with WorkGraph(
+    "AddMultiplyContextManager", inputs=spec.namespace(x=any, y=any, z=any)
+) as wg:
     the_sum = add(
         x=wg.inputs.x,
         y=wg.inputs.y,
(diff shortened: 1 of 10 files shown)


superstar54:
In the context_manager doc, I only updated the definition of the `wg.inputs`, but have not updated the text. Because the may text is not applicable anymore due to recent work (e.g., track graph-level inputs), thus, this doc needs a major refactoring, and we can do it in another PR.

e.g., 

```
#    Using the context manager paradigm allows for clear visualization of workflow inputs and their connections to tasks.
```
and the `Zone` approach can also be run inside the `Graph` task.

edan-bainglass:
> In the context_manager doc, I only updated the definition of the `wg.inputs`, but have not updated the text. Because the may text is not applicable anymore due to recent work (e.g., track graph-level inputs), thus, this doc needs a major refactoring, and we can do it in another PR.
> 
> e.g.,
> 
> ```
> #    Using the context manager paradigm allows for clear visualization of workflow inputs and their connections to tasks.
> ```
> 
> and the `Zone` approach can also be run inside the `Graph` task.

Thanks @superstar54. Indeed I saw that when I looked at it briefly. Thanks for the clear plan 🙂 I'll review now.

src/aiida_workgraph/workgraph.py
```diff
@@ -39,6 +40,7 @@ class WorkGraph(node_graph.NodeGraph):
     NodePool = TaskPool
     SocketPool = SocketPool
     platform: str = "aiida_workgraph"
+    type_mapping: dict = type_mapping
```
edan-bainglass, line 43: How do you use `type_mapping`?

docs/gallery/advanced/autogen/context_manager.py
```diff
@@ -109,15 +109,18 @@ def AddMultiply(x, y, z):
...
+    inputs=spec.namespace(
+        add=spec.namespace(
+            first=spec.namespace(x=any, y=any), second=spec.namespace(x=any, y=any)
+        ),
+        multiply=spec.namespace(factor=any),
+    ),
+    outputs=spec.namespace(
+        sums=spec.namespace(first=any, second=any, third=any), product=any
+    ),
+) as wg:
```
edan-bainglass, line 123: This is a lot 😅 Can we think of a way to reduce the amount of `spec.namespace` the user has to use by making so safe assumptions? Or at least provide a simpler API if no additional metadata if provided? For example, in `first=spec.namespace(x=any, y=any)`, the user isn't really making great use of a namespace. Can we automatically interpret `first=('x', 'y')` as an equivalent of the `any` case, making `spec.namespace` more of an opt-in feature?

edan-bainglass (approved):
Looks great @superstar54. Made a comment regarding possible in the near future considering a simpler API for those who aren't interested in the power of `spec.namespace` 🙂

docs/gallery/advanced/autogen/context_manager.py
superstar54: Yes, we can add this.
Note, if no inputs is specified, the `wg.inputs` will be set to be a dynamic namespace, which means one can add any input. In this case, 
```
wg.inputs = dict.fromkeys(["x", "y", "z"])
```
will still works.

src/aiida_workgraph/workgraph.py
superstar54: The type_mapping converts the Python type to a WorkGraph identifier. Currently, we only support simple types
e.g., `int -> workgraph.int`, etc

PR #641, opened 2025-08-20
Fix Graph task return namespace as outputs

### LINKED ISSUE #640: Wrong graph-level output when multiple outputs are returned in the while loop case
```python
from aiida_workgraph import task, WorkGraph, spec

load_profile()

@task()
def add(x, y):
    return x + y

@task.graph
def sum_to_n(n, total, N) -> spec.namespace(total=int, n=int):
    if n < N:
        total = add(n, total).result
        return sum_to_n(n + 1, total, N)
    else:
        return {'total': total, 'n': n}

N = 2
wg = sum_to_n.build_graph(1, 0, N)
wg.run()
print(wg.outputs)
```

Gives:
```python
TaskSocketNamespace(name='inputs', sockets=['result'])
```

superstar54:
This PR fixes:
- The generated WorkGraph should inherit the outputs from the graph task
- handle special case, where a graph task returns the whole output of a task.

Here is an example:
```python
from aiida_workgraph import task, spec

@task()
def add_multiply(x, y) -> spec.namespace(sum=int, product=int):
    return {"sum": x + y, "product": x * y}

@task.graph
def test_graph(x, y) -> spec.namespace(sum=int, product=int):
    # return the whole outputs of the task, which is a namespace
    return add_multiply(x, y)

wg = test_graph.build_graph(1, 2)
wg.run()
print(wg.outputs.sum)
print(wg.outputs.product)
```

## Fix the engine when the graph outputs contain `NodeLinkManager`

When the workgraph finishes, it will store the graph-level outputs to the AiiDA process node. This is done by:
```python
self.out_many(self.ctx._task_results["graph_outputs"])
```
 `self.ctx._task_results["graph_outputs"]` is a dict. Inside the dict, the value can be the output of a task. If the output is a namespace (`NodeLinksManager` attached to the AiiDA process node), for example
```python
In [1]: n = load_node(170588)

In [2]: n.outputs
Out[2]: <NodeLinksManager: Manager for outgoing RETURN links for node pk=170588>
```
We need to change it to a dict 
```python
{"sum": node1, "product": node2}
```
In this way, the `self.out_many` works. That's why we need this `clean_node_links_manager`.

### DIFF
--- a/src/aiida_workgraph/utils/__init__.py
+++ b/src/aiida_workgraph/utils/__init__.py
@@ -582,20 +582,20 @@ def serialize_socket_data(input_socket: Dict[str, Any]) -> None:
         input_socket["property"]["value"] = general_serializer(value)
 
 
-def resolve_node_link_manager_in_dict(data: Any) -> Any:
-    """Recursively resolve all NodeLinksManagers inside a dict."""
+def resolve_node_link_managers(data: Any) -> Any:
+    """Recursively resolve all NodeLinksManagers either in a dictionary or a NodeLinksManager."""
     if isinstance(data, dict):
         results = {}
         for key, value in data.items():
-            results[key] = resolve_node_link_manager_in_dict(value)
+            results[key] = resolve_node_link_managers(value)
         return results
     elif isinstance(data, orm.NodeLinksManager):
-        return node_link_manager_as_dict(data)
+        return convert_node_link_manager_to_dict(data)
     else:
         return data
 
 
-def node_link_manager_as_dict(
+def convert_node_link_manager_to_dict(
     node_link_manager: orm.NodeLinksManager,
 ) -> Dict[str, Any]:
     """Recursively convert a NodeLinksManager to a dictionary representation."""
@@ -603,7 +603,7 @@ def node_link_manager_as_dict(
     for name in node_link_manager._get_keys():
         item = node_link_manager._get_node_by_link_label(name)
         if isinstance(item, orm.NodeLinksManager):
-            data[name] = node_link_manager_as_dict(item)
+            data[name] = convert_node_link_manager_to_dict(item)
         else:
             data[name] = item
     return data
--- a/src/aiida_workgraph/utils/__init__.py
+++ b/src/aiida_workgraph/utils/__init__.py
@@ -580,3 +580,28 @@ def serialize_socket_data(input_socket: Dict[str, Any]) -> None:
         if isinstance(value, TaggedValue):
             value = value.__wrapped__
         input_socket["property"]["value"] = general_serializer(value)
+
+
+def clean_node_links_manager(data: Dict[str, Any]) -> Dict[str, Any]:
+    """Recursively clean a NodeLinksManager to a dictionary representation."""
+    results = {}
+    for key, value in data.items():
+        if isinstance(value, orm.NodeLinksManager):
+            results[key] = node_link_manager_to_dict(value)
+        else:
+            results[key] = value
+    return results
+
+
+def node_link_manager_to_dict(
+    node_link_manager: orm.NodeLinksManager,
+) -> Dict[str, Any]:
+    """Convert a NodeLinksManager to a dictionary representation."""
+    data = {}
+    for name in node_link_manager._get_keys():
+        item = node_link_manager._get_node_by_link_label(name)
+        if isinstance(item, orm.NodeLinksManager):
+            data[name] = node_link_manager_to_dict(item)
+        else:
+            data[name] = item
+    return data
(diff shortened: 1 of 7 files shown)


src/aiida_workgraph/utils/__init__.py
```diff
@@ -580,3 +580,28 @@ def serialize_socket_data(input_socket: Dict[str, Any]) -> None:
...
+) -> Dict[str, Any]:
+    """Convert a NodeLinksManager to a dictionary representation."""
+    data = {}
+    for name in node_link_manager._get_keys():
+        item = node_link_manager._get_node_by_link_label(name)
+        if isinstance(item, orm.NodeLinksManager):
+            data[name] = node_link_manager_to_dict(item)
+        else:
+            data[name] = item
+    return data
```
edan-bainglass: Can you provide some context on this? Also add it in the description of the PR 🙏

edan-bainglass (commented):
Thanks @superstar54. Can you please provide a comment in the PR description regarding the implementation? 🙏

src/aiida_workgraph/utils/__init__.py
superstar54: Yes. When the workgraph finishes, it will store the graph-level outputs to the AiiDA process node. This is done by:
```python
self.out_many(self.ctx._task_results["graph_outputs"])
```
 `self.ctx._task_results["graph_outputs"]` is a dict. Inside the dict, the value can be the output of a task. If the output is a namespace (`NodeLinksManager` attached to the AiiDA process node), for example
```python
In [1]: n = load_node(170588)

In [2]: n.outputs
Out[2]: <NodeLinksManager: Manager for outgoing RETURN links for node pk=170588>
```
We need to change it to a dict 
```python
{"sum": node1, "product": node2}
```
In this way, the `self.out_many` works. That's why we need this `clean_node_links_manager`.

edan-bainglass: Thanks 🙏 I have a few suggestions. I'll make them directly on the code.

src/aiida_workgraph/utils/__init__.py
```diff
@@ -580,3 +580,28 @@ def serialize_socket_data(input_socket: Dict[str, Any]) -> None:
         if isinstance(value, TaggedValue):
             value = value.__wrapped__
         input_socket["property"]["value"] = general_serializer(value)
+
+
+def clean_node_links_manager(data: Dict[str, Any]) -> Dict[str, Any]:
```
edan-bainglass:
```suggestion
def resolve_node_links_managers(data: Dict[str, Any]) -> Dict[str, Any]:
```

You aren't cleaning them (nothing is removed), just resolving them into their `dict` form.

src/aiida_workgraph/utils/__init__.py
```diff
@@ -580,3 +580,28 @@ def serialize_socket_data(input_socket: Dict[str, Any]) -> None:
...
+    """Recursively clean a NodeLinksManager to a dictionary representation."""
```
edan-bainglass:
```suggestion
    """Resolve `NodeLinksManager` instances into dictionary representation."""
```

This function is not the recursive one.

src/aiida_workgraph/utils/__init__.py
```diff
@@ -580,3 +580,28 @@ def serialize_socket_data(input_socket: Dict[str, Any]) -> None:
...
+            results[key] = node_link_manager_to_dict(value)
+        else:
+            results[key] = value
+    return results
+
+
+def node_link_manager_to_dict(
+    node_link_manager: orm.NodeLinksManager,
+) -> Dict[str, Any]:
+    """Convert a NodeLinksManager to a dictionary representation."""
```
edan-bainglass:
```suggestion
    """Recursively convert a `NodeLinksManager` to a dictionary representation."""
```

This is the recursive one.

edan-bainglass (approved):
LGTM! One comment regarding the use case...

```python
@task.graph
def test_graph(x, y) -> spec.namespace(sum=int, product=int):
    # return the whole outputs of the task, which is a namespace
    return add_multiply(x, y)
```

It would be nice if the user could avoid having to re-specify the identical output spec. Either we automate via detection (I believe you said this is not possible), or add an `expose_outputs` to `spec` and do

```python
@task.graph
def test_graph(x, y) -> spec.expose_outputs(add_multiply):
    # return the whole outputs of the task, which is a namespace
    return add_multiply(x, y)
```

edan-bainglass:
> LGTM! One comment regarding the use case...

Sorry, shouldn't have approved yet 😅 Please see my suggested changes 🙏

edan-bainglass (approved):
Once again, LGTM! Thanks for the function name updates 🙏

PR #645, opened 2025-08-27
Replace factory with TaskSpec

superstar54:
Remove the old factory class that creates the Node class on-the-fly. Now all dynamic tasks are created using the `SpecTask`.


edan-bainglass (approved):
LGMT!

PR #651, opened 2025-09-05
Bump aiida-pythonjob to 0.3.5

superstar54:
The aiida-pythonjob has been refactored recently, and a release 0.3.5 has been made.
The good thing is that, from aiida-workgraph side, there is not much change needed. 

- aiida-pythonjob to 0.3.5
- node-graph to 0.3.1

### DIFF
--- a/docs/gallery/tutorial/autogen/materials_science_ase.py
+++ b/docs/gallery/tutorial/autogen/materials_science_ase.py
@@ -258,7 +258,7 @@ wg.run()
 # %%
 # The result is an AiiDA Dict node. We access its content via the `.value` attribute.
 eos_result = wg.outputs.result.value
-print("Equation of state results for Cu: ", eos_result.value)
+print("Equation of state results for Cu: ", eos_result.get_dict())
 
 # %%
 # Visualize the EOS Provenance Graph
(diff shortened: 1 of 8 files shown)


docs/gallery/tutorial/autogen/materials_science_ase.py
```diff
@@ -258,7 +258,7 @@ def eos_workflow(atoms: Atoms, scales: list) -> dict:
 # %%
 # The result is an AiiDA Dict node. We access its content via the `.value` attribute.
 eos_result = wg.outputs.result.value
-print("Equation of state results for Cu: ", eos_result.value)
+print("Equation of state results for Cu: ", eos_result.get_dict())
```
superstar54, line 261: Note, aiida-pythonjob uses the `Dict`, `List` and `ArrayData` from aiida-core directly, instead of its own implementation that wraps them with a `value` attribute (because the wrapped class led to a query issue: https://github.com/aiidateam/aiida-pythonjob/issues/34).
So one has to use `get_dict`, `get_list`, but the good thing is that from `aiida-core 2.7.0`, `Dict` and `List` have a `value` attribute. So this change is needed only for the old aiida-core version.

edan-bainglass (approved):
LGTM!

PR #652, opened 2025-09-06
Check calling a task from inside another task

### LINKED ISSUE #648: Check calling a task inside another task instead of task.graph
One common mistake for users is that they may run a task inside another task, instead of using a task graph.

And it's hard to debug, because the function will run and return a Socket as a result.

Better to check and raise an error.

superstar54:
Fix: https://github.com/aiidateam/aiida-workgraph/issues/648

Raise an error and suggestions if the user calls a task inside the @task/@task.calcfunction/@task.workfunction body.

For the moment, checking inside the following task is not supported
- pythonjob task, because it runs on a remote computer
- monitor task, because it does not have its Process, instead it runs inside the WorkGraph process directly.


edan-bainglass (approved):
LGTM!

PR #653, opened 2025-09-06
Decorator accept `use_pickle` argument

### LINKED ISSUE #650: Add `allow_pickle` in the decorator
```python
@task(allow_pickle=True)
```
and pass this parameter to the aiida-pythonjob.

superstar54:
Needs tihs PR: https://github.com/scinode/node-graph/pull/85

This PR introduces a new `use_pickle` argument to the `@task` decorator, providing a convenient fallback mechanism for handling non-AiiDA and non-JSONable data nodes.

The `use_pickle` argument, when set to `True`, enables the engine to use `couldpickle` module as a fallback serialization method. This allows users to pass/return non-JSONable data types.

#### Example

```python

class MyData:
    """A custom, non-JSONable data class."""
    def __init__(self, value):
        self.value = value

@task(use_pickle=True)
def add_data(x, y):
    return x + y.value

with WorkGraph() as wg:
    wg.outputs.result = add(3, Data(4)).result
    wg.run()
```

This change makes AiiDA more accessible for new users and rapid development. For the production-level, robust workflows, the user is still recommended to provide a dedicated AiiDA data node for the custom data type.

### DIFF
--- a/docs/gallery/migration_from_aiida_core/autogen/workchain.py
+++ b/docs/gallery/migration_from_aiida_core/autogen/workchain.py
@@ -138,7 +138,7 @@ def sum_even_workgraph(N: int):
         wg.ctx.n = n_new.result
 
     # The 'result' step: define the final output of the graph.
-    wg.outputs.result = wg.ctx.total
+    return wg.ctx.total
 
 
 # %%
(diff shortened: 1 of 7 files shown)


edan-bainglass (approved):
LGTM! Once tests pass...

superstar54:
> LGTM! Once tests pass...

Thanks for the review. The tests passed, so I am merging this PR.

PR #676, opened 2025-09-14
Fix shelljob inputs

superstar54:
* Fix inputs for `prepare_shell_job_inputs`, use raw Python value instead of orm.Data for `command`, `resolve_command` and `arguments`.


edan-bainglass:
Hi @superstar54. Can you briefly explain why this was needed? 🙏

superstar54:
> Hi @superstar54. Can you briefly explain why this was needed? 🙏

Sorry that I missed this comment. `aiida-shell` provides `prepare_shell_job_inputs` helper, which is used in workgraph. Take the `command` input as an example, `prepare_shell_job_inputs` only accepts `str` and `orm.AbstractCode`, as shown [here](https://github.com/sphuber/aiida-shell/blob/64249d9e8e8b2880c98ed4df8633d50662645b0e/src/aiida_shell/launch.py#L86). However, the `command` of the task will be serialized to `orm.Str` if it's a `str`, which's why this change is needed.

PR #698, opened 2025-09-28
Use `call_link_label` to set task name

superstar54:
This PR provides a way to set the name of the task. The `call_link_label` is a built-in metadata input for the AiiDA process.
In WorkGraph, `call_link_label` is set to the same as the name of the task. In verse, when the user sets the `call_link_label`, we use it as the name of the task.
```python
sum1 = add(x, y, metadata={"call_link_label": "my_add"}).result
```

Also check that `metadata` is not used as an input argument of the function, because it is reserved by AiiDA.

## Docs
Add a new section: `Set task metadata`

Here is the preview: https://aiida-workgraph--698.org.readthedocs.build/en/698/howto/autogen/set_task_metadata.html

### DIFF
--- a/docs/gallery/howto/autogen/remote_job.py
+++ b/docs/gallery/howto/autogen/remote_job.py
@@ -59,6 +59,7 @@ print('\nResult: ', wg.outputs.result.value)
 # * **Outputs:** In addition to your function's return value (available as ``.result`` by default), the task provides comprehensive outputs, such as ``remote_folder`` (a reference to the job's directory on the remote computer).
 
 # %%
+# .. _pythonjob_metadata:
 # Prepare Python environment on remote computer
 # ---------------------------------------------
 #
@@ -90,8 +91,8 @@ def multiply(x, y):
 
 
 @task.graph
-def RemoteAddLocalMultiply(x: int, y: int, computer: str, metadata: Annotated[dict, add.inputs.metadata]) -> int:
-    the_sum = add(x=x, y=y, computer=computer, metadata=metadata).result
+def RemoteAddLocalMultiply(x: int, y: int, computer: str, add_metadata: Annotated[dict, add.inputs.metadata]) -> int:
+    the_sum = add(x=x, y=y, computer=computer, metadata=add_metadata).result
     return multiply(x=the_sum, y=4)  # this will run locally
 
 
@@ -99,7 +100,7 @@ wg = RemoteAddLocalMultiply.build(
     x=2,
     y=3,
     computer='localhost',
-    metadata=metadata,
+    add_metadata=metadata,
 )
 wg.run()
 
--- a/docs/gallery/howto/autogen/set_task_metadata.py
+++ b/docs/gallery/howto/autogen/set_task_metadata.py
@@ -1,0 +1,94 @@
+"""
+Set task metadata
+=================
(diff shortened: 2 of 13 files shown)


edan-bainglass:
Thanks @superstar54. Before I look into this, I first have a comment. Recent developments in node/work-graph shifted the package from using dictionaries to using utilities, e.g., `namespace`, `dynamic`, `select`, `meta`, etc. I am afraid this PR brings in another dictionary. These are prone to typing errors. Perhaps we can think up another mechanism?

Also, reminder of our discussion regarding the possibility to shift back to returning a task on call instead of the namespace, which would then allow operations on the task (which return a modified task), such as `.set_metadata`, or just `.set`, e.g. `add(x, y).set(name="my_add_task")`.

We've had such discussions in the past. They are tough to get out of 😅 But it is perhaps good to revisit. Happy to get on a call, or maybe we can add it to an MSD group meeting on open questions for the WorkGraph.

edan-bainglass (approved):
LGTM! Mind the comment though. I leave it up to you to decide if this should be merged in the short term.

superstar54:
Thanks @edan-bainglass for the comments! I merged this PR for the moment, but still keep it open for future changes on the API.


>  I am afraid this PR brings in another dictionary. These are prone to typing errors. Perhaps we can think up another mechanism?

This PR itself didn't introduce a new way to set metadata, because the `metadata` is a built-in input of the AiiDA process, thus, it's a namespace input of the task. The inputs will be validated in the same manner as other namespace inputs.
That's to say, without this PR, the user still sets the `metadata` of the process in this way. This PR reads this metadata and uses the `call_link_label` as the name of the task.

edan-bainglass:
> This PR itself didn't introduce a new way to set metadata, because the `metadata` is a built-in input of the AiiDA process, thus, it's a namespace input of the task. The inputs will be validated in the same manner as other namespace inputs. That's to say, without this PR, the user still sets the `metadata` of the process in this way. This PR reads this metadata and uses the `call_link_label` as the name of the task.

When I want to set the name of the task, I don't think of `call_link_label`. That is an internal artifact of AiiDA that likely should not be exposed to the user.

superstar54:
> When I want to set the name of the task, I don't think of call_link_label.

Agree. This is not intuitive for the users. In aiida-core, `call_link_label` is kind of the `name` of the called process, but it is not written as `name` explicitly.

Without modifying the aiida-core, in WorkGraph, we may create an alias "name" for it. This needs us to manually replace the `name` with `call_link_label` when setting the inputs of the task.

edan-bainglass:
> Agree. This is not intuitive for the users. In aiida-core, `call_link_label` is kind of the `name` of the called process, but it is not written as `name` explicitly.
> 
> Without modifying the aiida-core, in WorkGraph, we may create an alias "name" for it. This needs us to manually replace the `name` with `call_link_label` when setting the inputs of the task.

An alias would make users happy at the cost of a bit more code on our end. Reasonable. Though I'm not sure what you mean by "manually replace"

superstar54:
`manually replace` means `a bit more code on our end` :rofl:

edan-bainglass:
> `manually replace` means `a bit more code on our end` 🤣

Then reasonable 😅

PR #701, opened 2025-09-29
Fix loading process outputs

### LINKED ISSUE #700: Dynamic output is not loaded
Here is an example:

```python
from aiida_workgraph import task, dynamic
from typing import Annotated

load_profile()

@task
def generate_square_numbers(
    n: int,
) -> Annotated[dict, dynamic(int)]:
    """Generate a dict of square numbers. The number of outputs depends on 'n'."""
    return {f"square_{i}": i**2 for i in range(n)}

@task.graph
def SquareNumbersGenerator(n: int) -> Annotated[dict, generate_square_numbers.outputs]:
    return generate_square_numbers(n=n)

wg = SquareNumbersGenerator.build(5)
wg.run()
print(wg.tasks.generate_square_numbers.outputs)
```
It gives
```
TaskSocketNamespace(name='outputs', sockets=['_outputs', '_wait'])
```

When we inspect the outputs of the AiiDA process, it show outputs:
```
Outputs        PK  Type
---------  ------  ------
square_0   195666  Int
square_1   195667  Int
square_2   195668  Int
square_3   195669  Int
square_4   195670  Int
```

superstar54:
Fix #700 

- first,  use `resolve_node_link_managers` to get a dict representation of the node outputs
- then, set the dict as the output Socket value, and the Socket will handle dynamic/namespace automatically.

### DIFF
--- a/src/aiida_workgraph/task.py
+++ b/src/aiida_workgraph/task.py
@@ -137,7 +137,7 @@ class Task(GraphNode):
                 self.set_outputs_from_data_node(node)
 
     def set_outputs_from_process_node(self, node: aiida.orm.ProcessNode) -> None:
-        from aiida_workgraph.utils import get_nested_dict
+        from aiida_workgraph.utils import resolve_node_link_managers
 
         # if the node is finished ok, update the output sockets
         # note the task.state may not be the same as the node.process_state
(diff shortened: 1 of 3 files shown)


edan-bainglass (approved):
LGTM! Remind me briefly what `resolve_node_link_managers` does? 🙏

superstar54:
Convert NodeLinksManager to a dictionary representation.   For example, if `node. outputs` has `sum` and `product`, then it will return
```python
{"sum": value, "product": value}
```
It will resolve all NodeLinksManagers recursively. Thus can handle nested outputs.

PR #703, opened 2025-10-01
Update task and link names when using `call_link_label`

### LINKED ISSUE #702: Name inconsistent for tasks and links
In https://github.com/aiidateam/aiida-workgraph/pull/698, one can set the task name using `metadata`,
```python
sum1 = add(x, y, metadata={"call_link_label": "my_add"}).result
```
However, the name of the links is not updated, and results in an inconsistency.

superstar54:
Fix #702 

Update the names in the link and task collections after changing the name of the task.


edan-bainglass (approved):
LGTM!

PR #709, opened 2025-10-16
Support loading of a WorkGraph via its uuid

t-reents:
@superstar54 As discussed

I was just wondering whether we should rename the input argument `pk` as this is not fully consistent now. Nonetheless, it was already inconsistent before as a `ProcessNode` could also be provided for the `pk` argument. Something more general could be `identifier`. However, this would be backwards incompatible in case someone explicitly specifies the keyword. 

Let me not what you think @superstar54. I don't have a strong opinion at the moment, so happy to adjust it


superstar54 (approved):
LGTM!

PR #714, opened 2025-10-22
Fix function's scope for recursion

superstar54:
Always override the executor's globals to include the decorated function for recursion.


superstar54:
Hi @GeigerJ2 , this PR fixes the name issue.

PR #717, opened 2025-10-23
Attaching extra info to data nodes

superstar54:
## This PR
In the workgraph's annotation system, one can attach `extras` information for the socket. Originally, this was only used internally to store extra metadata. However, in principle, one can pass any JSON-serializable data into it. This is helpful if one wants to add additional (e.g., semantic) information to the data a task creates, such as a physical unit or a descriptive label. 

This information is stored in the database with the inputs and outputs `spec`, but not attached to the corresponding AiiDA data node. This PR populates the ``.extras`` attribute of the resulting AiiDA node with the `extras` of metadata the user provided. Thus, one can query the data using this extra info.

## Example

Let's see an example where we calculate the potential energy of a crystal structure and attach its unit to the output node.

```python
@task()
def calc_energy(atoms: Atoms) -> t.Annotated[float, meta(extras={"unit": "eV"})]:
    from ase.calculators.emt import EMT

    atoms.set_calculator(EMT())
    return atoms.get_potential_energy()

```

Later, one can fetch the unit directly from the result AiiDA node.
```
output_node = wg.outputs.result.value
energy_unit = output_node.base.extras.get('unit')
print(f'Calculated energy: {output_node.value} {energy_unit}')
```

### DIFF
--- a/docs/gallery/howto/autogen/annotate_inputs_outputs.py
+++ b/docs/gallery/howto/autogen/annotate_inputs_outputs.py
@@ -602,6 +602,47 @@ def UseMeta(
 
 
 # %%
+# Attaching extra info to data nodes
+# ----------------------------------
+# Sometimes you need to attach extra information to the data a task creates,
+# like a physical unit or a descriptive label.
+# You can do this directly in the output annotation by passing a dictionary
+# to ``meta(extras=...)``. This populates the ``.extras`` attribute of the
+# resulting AiiDA node, perfect for any custom, JSON-serializable metadata.
+# Let's see an example where we calculate the potential energy of a crystal
+# structure and attach its unit to the output node.
+#
+
+from ase import Atoms
+from ase.build import bulk
+
+
+@task()
+def calc_energy(atoms: Atoms) -> t.Annotated[float, meta(extras={'unit': 'eV'})]:
+    """Calculates the potential energy and attaches metadata to the output."""
+    from ase.calculators.emt import EMT
+
+    atoms.set_calculator(EMT())
+    return atoms.get_potential_energy()
+
+
+@task.graph()
+def MyWorkflow(atoms: Atoms):
+    """A simple workflow to run the energy calculation."""
+    return calc_energy(atoms).result
+
+
+wg = MyWorkflow.build(atoms=bulk('Cu'))
+wg.run()
+
+# %%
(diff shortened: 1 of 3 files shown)


superstar54:
Hi @edan-bainglass, this is one way to combine our annotation system with the AiiDA ORM data to allow users to provide extra info in the annotation and store it in the AiiDA database, thus enabling later inspection (e.g., queries).

edan-bainglass:
Thanks @superstar54. However, if you intended to use this as a gate for _semantic_ annotation, it may not be enough. Though this is probably useful, it is not really providing semantics in a way that is understood by tools that operate on semantics.

superstar54:
> tools that operate on semantics.

I am not familiar with all these tools, so I am always open to discussion. 

I guess semantic tools could work at both the workflow-definition level and the data level.

In the case of AiiDA, semantic info at the AiiDA workflow definition level is **not** easy in practice, as one cannot fetch a static knowledge graph with semantic info from a realistic (dynamic) AiiDA workflow.

At the data level, this is the power of AiiDA, because we keep track of full data provenance. Storing this extra info (can use the standard rule from the semantic community) along with the data is easy in practice and promotes interoperable data.

> if you intended to use this as a gate for semantic annotation, it may not be enough.

Agree. Here, we only show that it's possible to store extra info alongside the data through the standard annotation system. But what info to store depends on the semantic tools and use cases.

edan-bainglass:
> > tools that operate on semantics.
> 
> I am not familiar with all these tools, so I am always open to discussion.
> 
> I guess semantic tools could work at both the workflow-definition level and the data level.
> 
> In the case of AiiDA, semantic info at the AiiDA workflow definition level is **not** easy in practice, as one cannot fetch a static knowledge graph with semantic info from a realistic (dynamic) AiiDA workflow.
> 
> At the data level, this is the power of AiiDA, because we keep track of full data provenance. Storing this extra info (can use the standard rule from the semantic community) along with the data is easy in practice and promotes interoperable data.
> 
> > if you intended to use this as a gate for semantic annotation, it may not be enough.
> 
> Agree. Here, we only show that it's possible to store extra info alongside the data through the standard annotation system. But what info to store depends on the semantic tools and use cases.

Let's discuss in person 🙂

edan-bainglass (approved):
Thanks @superstar54. LGTM! We can discuss how to support semantics in a separate thread.

PR #727, opened 2025-11-11
Rolling task window

GeigerJ2:
__This works for our Sirocco use case__

Add support for topological-level-based rolling window submission,
which controls task submission based on their level in the workflow
dependency graph. This prevents overwhelming schedulers and provides
better resource management for large workflows.

Changes:

1. Add `extras` dict serialization to WorkGraph
   - Initialize `extras` dict in WorkGraph.__init__()
   - Serialize extras in to_dict() for database persistence
   - Deserialize extras in from_dict() to restore configuration
   - Enables passing custom metadata (like window_config) between
     workflow building and execution phases

2. Implement rolling window logic in TaskManager
   - Add window state tracking (min_active_level, max_allowed_level)
   - Load window configuration from WorkGraph.extras on initialization
   - Update window dynamically based on task completion
   - Gate task submission based on topological level
   - Only launcher tasks are gated; lightweight tasks always allowed (-> this is still hardcoded for sirocco)
   - Optional max_queued_jobs hard limit on concurrent tasks

__Sirocco stuff:__
Window behavior:
- window_size=0: Sequential execution (one level at a time)
- window_size=1: Levels N and N+1 can be active (default)
- window_size=N: N+1 consecutive levels can have active tasks
- Window advances automatically as lower-level tasks complete

The window configuration is provided by upstream workflow builders
via WorkGraph.extras['window_config'] with keys:
- enabled: bool - whether windowing is active
- window_size: int - number of levels in window
- max_queued_jobs: int|None - optional hard limit on concurrent jobs
- task_levels: dict - mapping of task names to topological levels

### DIFF
--- a/src/aiida_workgraph/engine/task_manager.py
+++ b/src/aiida_workgraph/engine/task_manager.py
@@ -44,6 +44,19 @@ class TaskManager:
         self.state_manager = TaskStateManager(ctx_manager, logger, process, awaitable_manager)
         self.action_manager = TaskActionManager(self.state_manager, logger, process)
 
+        # Initialize window state with defaults (will be loaded from WorkGraph context later)
+        self.window_config = {
+            'enabled': False,
+            'window_size': float('inf'),
+            'max_queued_jobs': None,
+            'task_levels': {},
+        }
+        self.window_state = {
+            'min_active_level': 0,
+            'max_allowed_level': float('inf'),
+        }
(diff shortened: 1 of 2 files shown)


superstar54:
Hi @GeigerJ2, could you provide an example to illustrate how this new feature works? I want to understand the use case better.

PR #732, opened 2025-12-03
Add notes about invalid task-level inputs

### LINKED ISSUE #486: Graph-level inputs: check order of operations
Reminder to check that when specifying a graph level input shared across tasks, also specifying the shared input at a task level indeed overrides the graph-level input.

ElliottKasoar:
Resolves #486

Adds a couple of examples of when you can no longer set task-level inputs (previously ignored, as discussed in the linked issue and https://github.com/scinode/node-graph/issues/131), including both implicit/explicitly linking tasks, and linking via graph-level inputs.

@superstar54, this seemed to be the most fitting section of the docs to me, as it talked most explicitly about linking and the example code fit well with what I wanted to describe, but I'm happy to move this/add more to the [section you suggested](https://aiida-workgraph.readthedocs.io/en/latest/advanced/autogen/context_manager.html#workflow-inputs-outputs) if you would like?

### DIFF
--- a/docs/gallery/concept/autogen/workgraph_concept.py
+++ b/docs/gallery/concept/autogen/workgraph_concept.py
@@ -42,6 +42,12 @@ wg.add_link(add1.outputs.result, add2.inputs.x)
 wg.to_html()
 
 # %%
+# .. note::
+#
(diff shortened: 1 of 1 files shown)


superstar54 (approved):
LGTM!

PR #765, opened 2026-02-27
Test aiida-core pre-release v2.8.0

agoscinski:
Test pass for v2.8.0 (see first commit). Py3.14 does not work because node-graph's pydantic version needs to be updated


alinelena:
now that 2.8.0 is out can we sync them?

ajjackson:
> Py3.14 does not work because node-graph's pydantic version needs to be updated

I don't follow this? node-graph specifies `"pydantic~=2.6"`, which should allow recent versions of Pydantic.

https://github.com/scinode/node-graph/blob/8d85e61c9024a6e04a9209bd677b94f159e073dd/pyproject.toml#L32

In the failed test run on Py3.12 we see that pydantic 2.11.9 was installed, which supports up to Python 3.14
https://github.com/aiidateam/aiida-workgraph/actions/runs/23037650794/job/66909196705?pr=765#step:6:159
https://pypi.org/project/pydantic/2.11.9/

GeigerJ2:
Hi all, we just merged #775 that loosens the `aiida-core` pin. I expanded the test matrix for CI to both 2.7 and 2.8, and everything passed, so I assume it's fine. Sorry, @agoscinski, I wasn't aware anymore of your PR here for the RC. Can this be closed, then?

PR #775, opened 2026-04-10
⬆️ Loosen `aiida-core` pin

### LINKED ISSUE #774: Loosen `aiida-core` version pin to `>=2.7.1`
## Loosen `aiida-core` version pin to `>=2.7.1`

`aiida-workgraph` pins `aiida-core~=2.7.1` in `pyproject.toml` (both main deps and docs extra), which means `>=2.7.1, <2.8.0`. This prevents installation alongside aiida-core `>=2.8` or dev versions, breaking CI/RTD builds and user installs.

Proposed: change to `>=2.7.1` (or `>=2.7.1,<3` if an upper bound is preferred).

The same applies to `aiida-pythonjob` (`~=0.5.1` / `~=0.5.2`).

We're integrating WorkGraph into the aiida-core tutorials ([PR #7205](https://github.com/aiidateam/aiida-core/pull/7205)). Module 3 teaches workflows using `@task.graph`, `shelljob`, and `Map`. The tutorial modules are executable MyST notebooks built by Sphinx, so `aiida-workgraph` must be installable alongside the current aiida-core during the docs build.

Adding `aiida-workgraph` to aiida-core's `tutorials` extra creates a circular dependency (workgraph depends on aiida-core, which is the package being built). With `uv`, this fails outright during resolution. With `pip`, it can resolve if aiida-core is already installed, but only if the version pin is satisfied. Since the aiida-core dev branch is `>=2.8`, the `~=2.7.1` pin rejects it either way.

Currently we work around this locally with an editable install from a branch that has the dependency restriction removed.

GeigerJ2:
Closes #774


pyproject.toml
GeigerJ2, line 1: Any reason it should be at least 2.7.**1**? Do we depend on anything from there?

GeigerJ2: doesn't matter anymore, we anyway yanked all `2.7.x` below `2.7.3`...

pyproject.toml
```diff
@@ -117,7 +117,7 @@ docs = [
   "myst-nb~=1.0.0",
   "nbsphinx",
   "furo",
-  "aiida-core~=2.7.1",
```
mbercx, line 120: Not introduced here, but: if it's already a core dependency, why is it also an optional one?

GeigerJ2: as we say in german: "doppelt haelt besser" :D 
jkjk, will remove. Good catch!

mbercx (approved):
LGTM! 🚀 

I can't immediately think of anything in `aiida-core` v2.8 that could break workgraph, but I don't really know the codebase. If the tests pass, we should be ok. If then something is still broken, it's an opportunity to improve our tests. Win-Win! :D

GeigerJ2:
> LGTM! 🚀
> 
> I can't immediately think of anything in `aiida-core` v2.8 that could break workgraph, but I don't really know the codebase. If the tests pass, we should be ok. If then something is still broken, it's an opportunity to improve our tests. Win-Win! :D

OK, tests have also passed with `aiida-core` 2.8 in the test matrix 🔥 I'd say this is good to go then!

t-reents:
@GeigerJ2 could we make a new release? I'm working on a workgraph related project that requires aiida-core 2.8.0 due to the support of newer numpy versions.

PR #776, opened 2026-04-13
Fix Map zone gather race for async process-type tasks

GeigerJ2:
**TL;DR** Stop cloning the pass-through gather_item; read the gathered values from the source clones directly

Mapping over an async process-type task (`CalcJob`, `WorkChain`, `@task.graph`) inside a `Map` zone could crash with a `KeyError` or hang the engine. At runtime a `Map` clones its body once per item into prefixed copies (`key_0_<task>`, `key_1_<task>`, ...), stored in `source_task.mapped_tasks[prefix]`.

The zone then gathers each clone's output through a pass-through `gather_item` node, and `update_map_task_state` read each item's result out of a *per-item clone of that `gather_item`*. For synchronous source tasks that's fine: each runs inline and is done before the engine moves on. For async ones, it races: the awaitable cascade can finish the mapped source tasks and reach the gather phase before `continue_workgraph` has scheduled the gather_item clones, so their results are missing (`KeyError`) or the clones sit in `PLANNED` (hang).

Two changes:
`generate_mapped_tasks` stops cloning the `gather_item`: it's a pure pass-through, so its clones only ever held copies, and the stray `PLANNED` clone was the hang. `update_map_task_state` instead reads each value directly from the mapped *source* task's clones (`source_task.mapped_tasks`), which is the same value one link upstream and is race-free, because a source task's results are written by `update_task_state` before the cascade reaches the gather.

The new read also passes `default=None`, which matters when a mapped item *fails*: `are_childen_finished` counts `FAILED`/`SKIPPED` as finished, so the zone still reaches the gather. This doesn't hide the failure. Like any failed task, the run finishes with the engine's existing exit status 302 (`is_finished_ok` False, message naming the failed task); `default=None` only stops the gather from raising a `KeyError` on the missing result, which on `main` excepted the whole engine and surfaced as `KeyError: 'sum1'`, burying the real cause (`key_1` raised a `ValueError`). The failed item lands as `None` in the gathered namespace, which is debatable: the honest value is arguably the failed process node itself (it carries the exit status and traceback), not a `None` that reads like data. That is a Map-semantics decision, separate from this race fix.

The existing map tests only map over synchronous tasks that always succeed, so they passed on `main` and never covered this. Added two to `tests/test_map.py` (a `@task.graph` source, and a failing item) that fail on `main` with `KeyError: 'sum1'` and pass here. Also dropped a stale commented-out block from `update_template_task_state`.

[#777](https://github.com/aiidateam/aiida-workgraph/pull/777) is stacked on this and exposes the gathered outputs to the client after `wg.run()`. I'd land this engine-side fix first; #777 re-touches `update_map_task_state` and can rebase onto it.

<details><summary>MWE</summary>

```python
from typing import Annotated
from aiida import orm
from aiida_workgraph import Map, WorkGraph, dynamic, namespace, task

@task()
def generate_data(n: int) -> Annotated[dict, namespace(data=dynamic(int))]:
    return {'data': {f'key_{i}': i for i in range(n)}}

@task()
def add(x, y):
    return x + y

@task.graph  # process-type: runs as its own sub-process (the async case)
def add_workflow(x, y) -> int:
    return add(x=x, y=y).result

@task()
def calc_sum(data: Annotated[dict, dynamic(orm.Int)]) -> float:
    return sum(data.values())

with WorkGraph('mwe') as wg:
    data = generate_data(n=3).data
    with Map(data) as map_zone:
        out1 = add_workflow(x=map_zone.item.value, y=1).result
        map_zone.gather({'sum1': out1})
    out3 = calc_sum(data=map_zone.outputs.sum1).result
    wg.run()

print(out3.value)   # main: KeyError: 'sum1' (engine EXCEPTED) -- here: 6
```

</details>

<details><summary>Why reading the source clone gives the same value</summary>

The gather link runs `source_clone.<from_socket>` -> `gather_item_clone.<to_socket>`, and `gather_item`'s executor is `return_input` (identity), so the value on the gather_item clone is literally the value already on the source clone.

|              | old (gather_item clone)        | new (source clone)                 |
|--------------|--------------------------------|------------------------------------|
| which clones | `gather_item.mapped_tasks`     | `source_task.mapped_tasks`         |
| read key     | `link.to_socket._name`         | `link.from_socket._scoped_name`    |
| accessor     | `d[key]`                       | `get_nested_dict(d, key, default=None)` |

Both loops use the same prefixes and write to the same `_task_results[name][link.to_socket._name]`. `_scoped_name` + `get_nested_dict` is used on the source side because a source output can be a nested namespace, whereas the gather_item input was always flat; for a flat socket they coincide.

</details>


elinscott (commented):
Thanks for this @GeigerJ2 ! The core fix is sound (much better than my fix, which dealt with rather than removed the race condition). I (read: Claude) did find one behavioral regression, however.

## Gathering a socket produced by a task outside the Map zone now silently yields an empty namespace

Since the gathered values are read from `from_task.mapped_tasks`, a source task outside the zone (which has no clones) contributes `{}`, the zone still reports FINISHED, and any downstream consumer of that output fails on a missing input. On the parent commit this worked: each per-item gather clone ran and fanned the outside value out per prefix, and the run finished with exit 0.

```python
with WorkGraph('probe_outside') as wg:
    data = generate_data(n=2).data
    outside = add(x=1, y=2).result  # outside the zone
    with Map(data) as map_zone:
        inner = add(x=map_zone.value, y=0).result
        map_zone.gather({'inner': inner, 'outside': outside})
    got = echo(data=map_zone.outputs.outside).result
    wg.run()
# parent commit: exit 0, echo receives {'key_0': 3, 'key_1': 3}
# this branch:  zone FINISHED but outputs.outside == {}, echo fails
#               (TypeError: echo() missing 1 required positional argument: 'data'), exit 302
```

This appears a little perverse (who would try to gather something outside of the `Map`?) but it's not unimaginable there would be a use case, so it would be good to fix. We could reject non-zone sources at `gather()` build time with a clear error, or fan the template task's single result out per prefix.

## Other points
- The new FAILED branch in `update_map_task_state` doesn't skip the zone's downstream tasks, unlike `on_task_failed`. With a task consuming `map_zone.outputs` after a failed iteration, the zone goes FAILED and the downstream task still runs, fails on its missing input, and pollutes the report and exit message (I get `['map_zone', 'calc_sum', 'key_1_maybe_fail']` where only the zone and the real failure belong). Adding `set_tasks_state(connectivity['child_node'][name], SKIPPED)` in that branch mirrors the ordinary failure path; a test with a task downstream of the failed zone would lock this in.
- `failed_prefixes` collects one entry per (link, prefix) pair, so two gather entries fed by the same failed source produce a duplicated message — `map_zone.gather({'a': out, 'b': out})` reports `no result from mapped item(s): key_1, key_1.` for me. `sorted(set(...))` fixes it.
- Treating SKIPPED source clones as failures looks right for the `on_task_failed` cascade, but it appears it would also fail the zone when a gather source sits inside a nested If zone whose condition is simply false for some item (previously gathered as `{prefix: None}`). I haven't tested this one — if If-inside-Map with gathered branch outputs is meant to be supported, it deserves a deliberate decision either way. Should we consider splitting SKIPPED into SKIPPED and UPSTREAM_FAILED?
- When a source clone fails, its `on_task_failed` flips the gather *template* from FINISHED to SKIPPED, since the clone's `child_node` connectivity still points at the template name now that no gather clone exists. (Doesn't affect behaviour, but ugly.)

src/aiida_workgraph/engine/task_state.py
```diff
@@ -291,49 +291,98 @@ def update_zone_task_state(self, name: str) -> None:
...
+        )
+        if errored_prefixes:
+            self.set_task_runtime_info(name, 'state', TaskState.FAILED)
+            # Skip the zone's downstream tasks, as `on_task_failed` does for an
+            # ordinary failure; otherwise they run on the missing gather output
+            # and pollute the report with their own consequent failures.
+            self.set_tasks_state(self.process.wg.connectivity['child_node'][name], TaskState.SKIPPED)
+            self.process.report(f'Task: {name} failed, no result from mapped item(s): {", ".join(errored_prefixes)}.')
             self.update_parent_task_state(name)
+            return
```
elinscott: This scans every clone in the zone, not just the ones feeding a gather link, so a failed side-effect task discards a complete gather (see MWE above).

Maybe key it on the gather sources instead: FAILED source clone → fail; SKIPPED source clone → fail only if a clone with the same prefix FAILED; otherwise None. That leaves unrelated body failures to the ordinary 302 path.

src/aiida_workgraph/engine/task_state.py
```diff
@@ -291,49 +291,98 @@ def update_zone_task_state(self, name: str) -> None:
...
+            results = {}
+            for prefix, clone in source_clones.items():
+                # A SKIPPED source (a false `If` branch) has no recorded result,
+                # so `default=None` gathers None for that item; a FINISHED source
+                # gathers its value.
+                results[prefix] = get_nested_dict(
+                    self.ctx._task_results[clone.name],
+                    link.from_socket._scoped_name,
+                    default=None,
+                )
```
elinscott, line 380: You say this default is for the SKIPPED `If` branch, but it can't tell that apart from a FINISHED source that simply didn't produce the socket.

```python
from typing import Annotated
from aiida_workgraph import Map, WorkGraph, dynamic, namespace, task

load_profile()


@task()
def generate_data(n: int) -> Annotated[dict, namespace(data=dynamic(int))]:
    return {"data": {f"key_{i}": i for i in range(n)}}


@task()
def sometimes(x) -> Annotated[dict, namespace(val=int)]:
    """Produce ``val`` for every item but one."""
    if x == 2:
        return {}
    return {"val": x * 2}


@task()
def consume(data: Annotated[dict, dynamic(int)]) -> int:
    print("CONSUME GOT:", data)
    return sum(int(v) for v in data.values() if v is not None)


with WorkGraph("missing-socket") as wg:
    data = generate_data(n=3).data
    with Map(data) as mz:
        mz.gather({"a": sometimes(x=mz.value).val})
    consume(data=mz.outputs.a)
    wg.run()
```

The zone finishes, the exit code is 0, nothing is reported, and the consumer prints

```
CONSUME GOT: {'key_0': 0, 'key_1': 2, 'key_2': None}
```

It would be nice to be able to distinguish between this and a skipped `If`.

src/aiida_workgraph/tasks/builtins.py
```diff
@@ -133,10 +133,28 @@ def gather_item_task(self) -> Task:
...
         """Collect per-entry results into the zone outputs, one namespace per name."""
         gather_item = self.gather_item_task
-        for name in sockets:
+        for name, socket in sockets.items():
+            source = socket._task
+            if not self._contains_task(source):
+                msg = (
+                    f"Map.gather(): output '{name}' comes from task '{source.name}', which is "
+                    f'outside the Map zone. A gather source must be produced inside the zone '
+                    f"(one value per iteration); move '{source.name}' into the `with Map(...):` block."
```
elinscott: When the source is a graph-level input this message is misleading:

```python
from typing import Annotated, Any
from aiida_workgraph import Map, WorkGraph, dynamic, namespace, spec, task

load_profile()


@task()
def generate_data(n: int) -> Annotated[dict, namespace(data=dynamic(int))]:
    return {"data": {f"key_{i}": i for i in range(n)}}


@task()
def add(x, y):
    return x + y


with WorkGraph("graph-input-gather", inputs=spec.namespace(shift=Any)) as wg:
    wg.inputs.shift = 1
    data = generate_data(n=3).data
    with Map(data) as mz:
        good = add(x=mz.value, y=wg.inputs.shift).result
        mz.gather({"a": good, "ext": wg.inputs.shift})
```

```
ValueError: Map.gather(): output 'ext' comes from task 'graph_inputs', which is
outside the Map zone. A gather source must be produced inside the zone (one value
per iteration); move 'graph_inputs' into the `with Map(...):` block.
```

Rejecting is right — a graph-level input is one value, not a per-iteration one — but the advice is wrong.

src/aiida_workgraph/engine/task_manager.py
```diff
@@ -472,6 +472,15 @@ def generate_mapped_tasks(self, zone_task: Task, prefix: str) -> None:
...
         for child_task in child_tasks:
+            # The gather_item task is a pure pass-through aggregator
+            # (executor=return_input); the map zone reads directly from the
+            # mapped source tasks in `update_map_task_state`, so cloning
+            # gather_item would just create unused clones. Skipping the
+            # clone also avoids a race where, for async process-type source
+            # tasks (CalcJob, WorkChain, @task.graph), the gather_item
+            # clones stay PLANNED and hang the engine's finalize path.
+            if self.process.wg.tasks[child_task].identifier == 'workgraph.gather_item':
+                continue
```
elinscott, line 484: With gather_item no longer cloned, `map_info['children']` no longer lists it and the source→gather edges are dropped by the `link.to_task.name in new_tasks` filter. `map_info` only feeds the GUI as far as I can see — might affect how the web UI draws these edges...

elinscott (commented):
Thanks @GeigerJ2 !

Four of my five points addressed. The fifth is the cosmetic gather-template flip (a failed source clone still flips the gather template FINISHED→SKIPPED).

However, I still found a few things that need fixing. The headline issue is that the new changes mean that the FAILED scan walks every clone, so a failing unrelated body task now makes the zone fail with a misleading error message. (On `main` this gave more rational behaviour: that case gathered completely and the run exited 302 for the failed task.)

```python
from typing import Annotated
from aiida_workgraph import Map, WorkGraph, dynamic, namespace, task

load_profile()


@task()
def generate_data(n: int) -> Annotated[dict, namespace(data=dynamic(int))]:
    return {"data": {f"key_{i}": i for i in range(n)}}


@task()
def add(x, y):
    return x + y


@task()
def side_effect(x):
    if x == 1:
        raise ValueError("boom")
    return x


@task()
def consume(data: Annotated[dict, dynamic(int)]) -> int:
    return sum(int(v) for v in data.values())


with WorkGraph("body-failure") as wg:
    data = generate_data(n=3).data
    with Map(data) as mz:
        good = add(x=mz.value, y=1).result
        side_effect(x=mz.value)      # fails for key_1; output never gathered
        mz.gather({"a": good})
    consume(data=mz.outputs.a)       # main: ran, returned 6.  here: SKIPPED
    wg.run()
```

The report then says `no result from mapped item(s): key_1`, but every gathered item has a result. See the inline comment below for more details.

Other smaller issues are also listed as inline comments.

Also note in the meantime I've dropped the overlapping fix from #797

src/aiida_workgraph/engine/task_state.py
GeigerJ2: Agreed, and there's a third case that reads the same `None`: a FINISHED source that returns no value for the socket. Telling them apart is really the resilient-`Map` / per-item-status design, so I've written it up separately (#TBD) rather than force it into this race fix.

src/aiida_workgraph/engine/task_state.py
GeigerJ2: Good catch, fixed. The check is keyed on the gather sources now: a FAILED source fails the zone, a SKIPPED source only if a same-prefix clone also FAILED, otherwise `None`. So an unrelated body-task failure keeps the gather complete and just surfaces via exit 302. Added `test_map_unrelated_body_failure_does_not_fail_zone`.

src/aiida_workgraph/tasks/builtins.py
GeigerJ2: Fixed, thanks. Shortened it and dropped the "move it into the zone" advice; it now just says the source is outside the zone (one value, not per-iteration) and gives both routes, move a per-iteration task in, or use a shared value like a graph input directly.

src/aiida_workgraph/engine/task_manager.py
GeigerJ2: Fixed. Restored the source→gather_item edges in `map_info` as template edges (the shape the UI expands per prefix) and put `gather_item` back in `children` so they resolve. Added `test_map_info_keeps_gather_edges`.

elinscott (approved):
Looks good to me. One very minor, non-blocking thing: `Map.gather` raises inside its loop, _after_ earlier names have had their input/output specs added, so if someone builds the graph within a try/except the `ValueError` leaves a half-built zone which then fails upon rebuilding due to sockets already existing. Validating all sockets before adding any specs is a two-line reorder, but what sort of madman would try to catch a build-time error and keep going?

GeigerJ2:
Good catch, fixed it, thanks! Added you as co-author, as reviewing is the main job nowadays :)

PR #777, opened 2026-04-14
Expose Map zone outputs to the client after `wg.run()`

GeigerJ2:
Zone/Map tasks have no AiiDA process node (`pk=None`), so
`Task.update_state` could never populate their outputs on the client
side. After `wg.run()`, `map_zone.outputs.<name>` returned an empty dict
even though the engine had the correct gathered results internally.

Three changes:

- `gather()` now creates dynamic output namespaces so the client can
  assign per-prefix keys after the run.

- `update_map_task_state` persists the gathered result node PKs in
  `task_map_info[name]['result_pks']` on the process node.

- `WorkGraph.update()` reads those PKs back and populates the zone
  task's output sockets via `_populate_zone_outputs`.


GeigerJ2:
@coderabbitai full review

GeigerJ2:
@coderabbitai full review

GeigerJ2:
@coderabbitai full review

GeigerJ2:
@coderabbitai full review

PR #787, opened 2026-06-22
👌 Validate task names as AiiDA link labels on `build`

### LINKED ISSUE #784: An invalid task-name link label fails silently at run time instead of at build time
A `@task` / `@task.graph` name becomes an AiiDA link label, which must match `[A-Za-z0-9_]+` and may not start with `_`. An invalid name is not caught at build time. At run time the task fails inside the engine (`validate_link_label`: "cannot start with an underscore"), but `WorkGraph.run()` does not raise: the workgraph is reported as finished, the offending task is marked failed and skipped, and it simply produces no output.

## MWE
Requires a profile (the failure is at run time).

```python

@task
def _hidden() -> dict:          # leading underscore -> invalid link label
    return {"ran": True}

@task.graph
def top():
    _hidden()

top.build().run()
# run() does not raise; log shows:
#   ValueError: invalid link label `_hidden`: cannot start with an underscore
#   WorkGraph finished, but tasks: ['_hidden'] failed.
# `_hidden` produced no output.
```

GeigerJ2:
A task name becomes the process's `call_link_label`, so it has to be a valid link label.

**Alternatives considered:** auto-fixing the name (prepend `fn_`, strip underscores, ...) would hand the user a name they never wrote, which is then what shows up in their provenance and `wg.tasks.<name>` lookups, and no transform is obviously right anyway (stripping collides `_a`/`a`, prefixing is arbitrary). Such a name is almost always an oversight, so failing fast with a fix hint beats persisting a renamed node, and anyone who does want a custom name still sets one explicitly via `name=` or `call_link_label`.

Closes #784.

### DIFF
--- a/src/aiida_workgraph/task.py
+++ b/src/aiida_workgraph/task.py
@@ -215,8 +215,15 @@ class TaskHandle(BaseHandle):
         outputs = super().__call__(*args, **kwargs)
         # if "metadata.call_link_label" is set, use it as the name of the task
         if outputs._task.inputs.metadata.call_link_label.value is not None:
+            from aiida_workgraph.utils import _validate_task_name
+
             graph = outputs._graph
-            outputs._task.name = outputs._task.inputs.metadata.call_link_label.value
+            new_name = outputs._task.inputs.metadata.call_link_label.value
+            # `call_link_label` overrides the (already validated) function-derived name, so it
+            # must be validated too; otherwise an invalid override slips past `add_task` and
+            # only fails silently inside the engine at run time (see issue #784).
+            _validate_task_name(new_name, source='call_link_label')
+            outputs._task.name = new_name
             # update the names of tasks and links collections in the graph
             graph.tasks._items = {task.name: task for task in graph.tasks._items.values()}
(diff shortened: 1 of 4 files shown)


elinscott (approved):
This fixes my test in `aiida-workflow-wishlist`. The implementation looks clean, if a little comment-heavy.

GeigerJ2:
Thanks for the review, @elinscott!

> This fixes my test in `aiida-workflow-wishlist`.

I guess rather than fix, fail early and clearly, no?

> The implementation looks clean, if a little comment-heavy.

Yeah, that's my agent ignoring my _be concise_ CLAUDE.md instruction :D though, I tend to keep such verbose comments / commit msgs / PR bodies if they outline the thought process I went through, so it's also clear for future devs. In this case, actually noticing and accounting for the three places where the validation has to be done took a few iterations.

elinscott:
Thanks, nice to see this merged!

> > This fixes my test in `aiida-workflow-wishlist`.
> 
> I guess rather than fix, fail early and clearly, no?

"Fix" in the sense that the test passes when the code fails loudly :wink:

PR #788, opened 2026-06-23
🐛 Fix `@task.graph` failing under `from __future__ import annotations`

### LINKED ISSUE #783: from __future__ import annotations breaks a dynamic-namespace graph task at run time
A deferred `@task.graph` whose signature uses a dynamic-namespace annotation fails at run time under `from __future__ import annotations`. The engine re-infers the graph task's signature while running it, and the stringised annotation cannot be resolved: `NameError: name 'Annotated' is not defined`. The task is reported as failed mid-run. Dropping the future import avoids it, but PEP 563 is the modern default.

## MWE

Requires a profile (the failure is at run time, not build).

```python
from __future__ import annotations
from typing import Annotated
from aiida_workgraph import dynamic, namespace, task

@task
def source() -> Annotated[dict, namespace(data=dynamic(int))]:
    return {"data": {"k1": 1, "k2": 2}}

@task
def record(v) -> dict:
    return {"v": int(v)}

@task.graph
def fan(data: Annotated[dict, dynamic(int)]):
    for _k, v in data.items():
        record(v=v)

@task.graph
def top():
    fan(data=source().data)

top.build().run()
# at run time, task `fan` fails with:
#   NameError: name 'Annotated' is not defined
```

## Root cause

The engine re-infers the `fan` graph task's signature at run time:

```
WorkGraphEngine ... runs GRAPH task `fan`
  -> aiida_workgraph/tasks/function_task.py: build_callable_TaskSpec
  -> node_graph/socket_spec.py: infer_specs_from_callable
  -> node_graph/socket_spec.py: _safe_type_hints
  -> typing.get_type_hints(func, include_extras=True)
  -> NameError: name 'Annotated' is not defined
```

node-graph itself handles PEP 563: its own engine runs the equivalent dynamic-namespace graph (closure included) without error, because it keeps the live function and never re-resolves its hints in a context that lacks the imports. The failure is specific to the run-time re-inference here.

GeigerJ2:
`@task.graph` with a dynamic-namespace signature fails at run time under `from __future__ import annotations`: `NameError: name 'Annotated' is not defined`, and the task fails mid-run.

Under PEP 563 the annotation is just a string that `get_type_hints` resolves lazily against the function's globals, but cloudpickle drops names used only inside stringized annotations from the unpickled callable. `GraphTask.execute` re-decorated that callable just to register a recursion handle, which re-inferred the signature and tripped over the missing `Annotated`. The fix builds the handle from the existing `self.spec` instead, so nothing re-infers the signature at run time. Includes a regression test.

MWE provided in the corresponding issue raised by @elinscott.

### DIFF
--- a/src/aiida_workgraph/tasks/function_task.py
+++ b/src/aiida_workgraph/tasks/function_task.py
@@ -1,4 +1,6 @@
 from __future__ import annotations
+import contextlib
+import typing
 from typing import Callable, List, Optional, Type, Dict, TYPE_CHECKING
 from aiida_workgraph.socket_spec import (
     from_aiida_process,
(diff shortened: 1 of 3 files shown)


elinscott (approved):
This fixes the issue for me.

I see you re-marked the PR as "draft"; feel free to re-request my review if desired.

PR #790, opened 2026-06-23
♻️ `engine`: use enums for task state and action

### LINKED ISSUE #789: Replace stringly-typed values with `Enum`/`Literal`
The engine encodes a few closed value sets as bare strings checked against hand-written `in [...]` lists copy-pasted across files: no single source of truth, a typo silently never matches, and a new member has to be remembered in every list. The aiida-workgraph-owned ones:

| domain | values | uses | suggested |
|---|---|---|---|
| task state | `PLANNED` `RUNNING` `CREATED` `READY` `FINISHED` `FAILED` `SKIPPED` `MAPPED` | ~92, 13 files | `Enum` (+ `TERMINAL`/`is_terminal()`) |
| task action | `PAUSE` `PLAY` `KILL` `SKIP` `RESET` | ~15 | `Enum` |
| runtime-info keys | `process` `state` `action` `execution_count` `map_info` | get/set dispatch | `Literal` |

Task state is the highest-value: the duplicated `['FINISHED','SKIPPED','FAILED']` terminal check plus ad-hoc `MAPPED` handling is what made [#776](https://github.com/aiidateam/aiida-workgraph/pull/776) subtle. `action` already has a latent bug, some sites compare `.upper() == 'RESET'`, others raw `== 'RESET'`, so a wrong-cased value matches in some places but not others.

Proposal: `TaskState(str, Enum)` / `TaskAction(str, Enum)` plus a runtime-key `Literal`, with helpers like `TaskState.TERMINAL`. The `str` mixin keeps members equal to and JSON-serialising as today's strings, so persisted process-node state is unchanged and the migration is mechanical and incremental. Bonus: autocomplete, typo protection, `assert_never` exhaustiveness.

The `_wait`/`_outputs` builtin socket names belong to node-graph (which already defines constants for them); tracked in a sibling node-graph issue.

GeigerJ2:
Task state and action were bare strings checked against copy-pasted `in [...]` lists. This PR adds `aiida_workgraph/enums.py` (`TaskState`/`TaskAction` as `str`-mixed enums + a `RuntimeInfoKey` `Literal`), migrates every engine call site, and unifies the terminal check into `TERMINAL_TASK_STATES` / `is_terminal`.

Behaviour-preserving: the `str` mixin keeps members equal to and JSON-serialising as the same bare strings, so persisted process-node state is byte-identical and existing string comparisons / `wg.wait(...)` keep working. It also fixes the latent action casing bug — normalise once in `apply_task_actions` (validate to `TaskAction`, dispatch via `match`), drop the scattered `.upper()`, and an unknown action now raises instead of being silently ignored.

Scope: `node.process_state` comparisons stay strings (aiida-core's domain), the dead `WAITING`/`PAUSED` are dropped from `kill_tasks`, and `enums.py` is importable but not in the top-level `__all__`. Typing follow-ups (`TaskType` enum, removing the runtime-info dispatch, `TypedDict`s, `Optional` sentinels, a node-graph sibling) will be filed as separate issues.

Closes [#789](https://github.com/aiidateam/aiida-workgraph/issues/789).

### DIFF
--- a/src/aiida_workgraph/engine/task_actions.py
+++ b/src/aiida_workgraph/engine/task_actions.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
-from typing import Callable, Dict
+from typing import Callable
 
 from typing_extensions import assert_never
 
-from aiida_workgraph.enums import TaskAction, TaskState
+from aiida_workgraph.enums import TaskAction, TaskActionMessage, TaskState
 
 
 class TaskActionManager:
@@ -21,15 +21,14 @@ class TaskActionManager:
         self.logger = logger
         self.process = process
 
-    def apply_task_actions(self, msg: Dict) -> None:
+    def apply_task_actions(self, msg: TaskActionMessage) -> None:
         """
         Apply task actions to the workgraph based on user or external messages.
 
-        :param msg: { "action": <str>, "tasks": <List[str]> }
         :raises ValueError: if the message carries an unknown task action.
         """
         action = TaskAction(msg['action'].upper())
-        tasks: list[str] = msg['tasks']
+        tasks = msg['tasks']
         self.process.report(f'Action: {action}. Tasks: {tasks}')
 
         # Each action maps to a callable that takes a single task name.
--- a/src/aiida_workgraph/engine/task_actions.py
+++ b/src/aiida_workgraph/engine/task_actions.py
@@ -1,5 +1,9 @@
 from __future__ import annotations
-from typing import Dict
+from typing import Callable, Dict
+
+from typing_extensions import assert_never
+
+from aiida_workgraph.enums import TaskAction, TaskState
 
 
 class TaskActionManager:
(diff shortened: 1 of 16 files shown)


src/aiida_workgraph/engine/task_actions.py
```diff
@@ -20,31 +24,35 @@ def __init__(self, state_manager, logger, process):
     def apply_task_actions(self, msg: Dict) -> None:
```
elinscott: Any reason why msg should remain a dict rather than separate `action` and `tasks` args? Would get tighter typechecking with the latter and would make the `assert_never` redundant

(Or, even better, `class Message(TypedDict)`)

src/aiida_workgraph/engine/task_manager.py
```diff
@@ -49,8 +50,9 @@ def get_task(self, name: str):
         """Get task from the context."""
         task = self.process.wg.tasks[name]
         task.set_input_resolver(self.get_socket_value)
-        action = self.state_manager.get_task_runtime_info(name, 'action').upper()
-        task.action = action
+        # The action stored on the node is always a canonical TaskAction value (or
+        # '' for none), so no case normalisation is needed on read.
```
elinscott: Remove redundant comment

src/aiida_workgraph/engine/task_state.py
```diff
@@ -25,39 +26,40 @@ def __init__(self, ctx_manager, logger, process, awaitable_manager):
...
+                value = self.process.node.get_task_process(name)
+                return deserialize_safe(value) if value else None
+            case 'state':
+                return self.process.node.get_task_state(name)
+            case 'action':
+                return self.process.node.get_task_action(name)
+            case 'execution_count':
+                return self.process.node.get_task_execution_count(name)
+            case _:
+                raise ValueError(f'Invalid key: {key}')
```
elinscott, line 43: Doesn't `key`'s type make the ValueError redundant (provided we're good and check typing)?

src/aiida_workgraph/utils/control.py
```diff
@@ -6,32 +6,38 @@
...
-    action: str = 'pause',
+    action: TaskAction = TaskAction.PAUSE,
 ):
     """Send task action to Process."""
 
     controller = get_manager().get_process_controller()
-    message = {'intent': 'custom', 'catalog': 'task', 'action': action, 'tasks': tasks}
+    # Send the canonical action value as a plain string; the engine re-validates it
+    # into a TaskAction on receipt.
+    message = {'intent': 'custom', 'catalog': 'task', 'action': str(action), 'tasks': tasks}
```
elinscott: How about `class Message(TypedDict)`?

tests/test_enums.py
```diff
@@ -0,0 +1,64 @@
...
+
+from aiida_workgraph.enums import TERMINAL_TASK_STATES, TaskAction, TaskState
+
+
+@pytest.mark.parametrize('enum_cls', [TaskState, TaskAction])
+def test_str_and_format_drop_class_prefix(enum_cls):
+    """``__str__``/f-strings must yield the bare value on every supported Python
+    version, so report messages stay e.g. ``Action: RESET`` (test_workgraph)."""
+    member = next(iter(enum_cls))
+    assert str(member) == member.value
```
elinscott, line 18: All three necessary?

elinscott (approved):
Just a few minor non-blocking suggestions

src/aiida_workgraph/engine/task_actions.py
GeigerJ2: Exactly what I'm talking about :) good catch, done!

src/aiida_workgraph/engine/task_state.py
GeigerJ2: > provided we're good and check typing

unfortunately, we are not (yet) ^^ but soon: https://github.com/aiidateam/aiida-workgraph/pull/795

tests/test_enums.py
GeigerJ2: agreed, trimmed to two

GeigerJ2:
Thanks for the review, @elinscott! All applied :)

Issue #786, opened 2026-06-22
Inconsistency when .value is required


elinscott:
Inside a deferred `@task.graph` body...

- a scalar input arrives as a `TaggedValue` over an `orm` node (e.g. `orm.Str`). Naive use raises and you must reach through `.value`
- A dict input in the same position arrives as a plain dict you can subscript directly.

The asymmetry is surprising — there is no signal that one input needs `.value` and the other does not.

This is possbly related to the same boundary-promotion seen in #780 (scalars promoted to `orm` types across a `@task.graph` boundary), here for a direct scalar input rather than a dataclass field. It is not a node-graph issue: in pure node-graph the scalar arrives as a `TaggedValue` over a plain `str` and `int(label)` works without `.value`. The promotion to `orm.Str` is added by aiida-workgraph.

## MWE

```python
from aiida_workgraph import task

@task
def make_label() -> str:
    return "777"

@task
def make_dict() -> dict:
    return {"n": 5}

@task
def sink(scalar, n) -> dict:
    return {"scalar": scalar, "n": n}

@task.graph
def inner(label, d):
    # int(label) raises -- label is a TaggedValue over orm.Str, not a str
    sink(scalar=int(label.value), n=d["n"])   # scalar needs .value; dict subscripts directly

@task.graph
def top():
    inner(label=make_label().result, d=make_dict().result)

top.build().run()
```

## Wish

A scalar input should behave like its plain Python value in a `@task.graph` body (no `.value`), consistent with how a dict input is delivered.

PR #792, opened 2026-06-23
👌 `Map`: replace `z.item` with `z.value`/`z.key`

### LINKED ISSUE #785: It's unclear what you iterate over in a Map zone
One usability problem with a `Map` zone is that it is not immediately obvious what you are iterating over. The name `Map`, and the per-element handle `z.item` suggest *sequence* iteration — you get elements/values, the way `map(f, seq)` or a list comprehension hands you each value. But the source is actually always a dynamic namespace — i.e. a dict — and iteration is over keyed entries, never a bare sequence. Nothing in the API signals which model applies, so users guess wrong (at least, I did!)

In fact there is no sequence path at all: a task that returns a bare list to a dynamic output is rejected at runtime (`parse_outputs([...], dynamic(...)) -> None`); the source must be a dict. So keys always exist — even when the data is logically a list and the keys are synthetic indices like `"0"`, `"1"`.

The per-element handle compounds the confusion: `z.item` is not the element — it is the `map_item` task's outputs namespace, with two ports:

- `z.item.value` — the value for that element
- `z.item.key` — the source key for that element (your own key, e.g. `"i1"`/`"i2"` below — not internal machinery)

Passing the bare `z.item` raises at build time:

```
ValueError: You try to link a top-level output socket outputs without a parent.
```

So a user must discover the `.value` / `.key` ports (possibly by trial and error) — with no signal that the thing is dict-shaped rather than list-shaped, and no clear, Python-aligned way to say "give me the values" vs "the keys" vs "the pairs".

Maybe this will be less of an issue as the docs expand, but without docs to guide me I didn't find the current implementation intuitive. 

## MWE

```python
from typing import Annotated
from aiida_workgraph import Map, dynamic, task

@task
def make_items() -> Annotated[dict, dynamic(dict)]:
    return {"i1": {"n": 7}, "i2": {"n": 99}}

@task
def sink(dct) -> dict:
    return {"n": dct["n"]}

@task.graph
def top():
    with Map(make_items()) as z:
        sink(dct=z.item)          # fails; only z.item.value works -- quite verbose

top.build()
# ValueError: You try to link a top-level output socket outputs without a parent.
```

## Suggestions
This is quite an open-ended design question. Alterantives to the current setup which I think would be more intuitive (or at least follow vanilla `python` more closely):
- create `z.value` and `z.key` (even if internally they just point to `z.item.value` and `z.item.key`)
- `z.item` being a tuple (to match python's `.items()`)

GeigerJ2:
A `Map` zone always iterates a dynamic namespace (a dict), but the per-element handle `z.item` never made that obvious: it was a namespace whose `.value`/`.key` ports you had to discover, passing bare `z.item` raised a cryptic node_graph error, and the name suggested sequence iteration (#785).

This replaces `z.item` with two explicit handles on the zone, `z.value` for the current entry's value and `z.key` for its key:

```python
with Map(make_items()) as z:
    sink(dct=z.value)      # was: z.item.value
    label(text=z.key)      # was: z.item.key
```

Since there is no stable release, `z.item` is dropped fully rather than deprecated. Also drops an unused `placeholder` parameter from the `Map` context manager.

The generic node_graph "top-level output socket without a parent" message (seen when linking any bare namespace into a leaf input) is left as a separate upstream follow-up.

Closes #785

### DIFF
--- a/src/aiida_workgraph/tasks/builtins.py
+++ b/src/aiida_workgraph/tasks/builtins.py
@@ -91,7 +91,12 @@ class If(Zone):
 
 
 class Map(Zone):
-    """Map"""
+    """Run a set of tasks once per entry of a dynamic-namespace (dict) source.
+
+    A Map iterates keyed entries, not a bare sequence. While building the body,
+    ``value`` and ``key`` are placeholders for the current entry; the engine
+    clones the body once per entry and binds them.
+    """
 
     _default_spec = TaskSpec(
         identifier='workgraph.map_zone',
@@ -107,14 +112,22 @@ class Map(Zone):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-    @property
-    def item(self):
+    def _map_item_task(self) -> Task:
+        """Return the single ``map_item`` child, creating it on first access."""
         for child in self.children:
             if child.identifier == 'workgraph.map_item':
-                return child.outputs
-        # create a child map_item_task if it does not exist
-        map_item_task = self.add_task('workgraph.map_item')
-        return map_item_task.outputs
+                return child
+        return self.add_task('workgraph.map_item')
+
+    @property
+    def value(self) -> BaseSocket:
+        """Placeholder for the current entry's value while iterating the source."""
+        return self._map_item_task().outputs.value
+
+    @property
+    def key(self) -> BaseSocket:
+        """Placeholder for the current entry's key while iterating the source."""
+        return self._map_item_task().outputs.key
 
     @property
     def gather_item_task(self) -> Task | None:
@@ -124,7 +137,8 @@ class Map(Zone):
         gather_item = self.add_task('workgraph.gather_item')
         return gather_item
 
-    def gather(self, sockets: Dict[str, BaseSocket]) -> None:
+    def gather(self, sockets: Dict[str, BaseSocket]) -> BaseSocket:
+        """Collect per-entry results into the zone outputs, one namespace per name."""
         gather_item = self.gather_item_task
         for name in sockets:
             gather_item.add_input_spec('workgraph.any', name=name)
--- a/docs/gallery/advanced/autogen/context_manager.py
+++ b/docs/gallery/advanced/autogen/context_manager.py
@@ -489,9 +489,8 @@ wg.generate_provenance_graph()
 #   We welcome your feedback on its functionality.
 #   **ctx** does not work inside a ``Map`` zone yet.
 #
-# The ``Map`` context manager works similarly to Python's built-in ``map`` function.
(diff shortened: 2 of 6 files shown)


elinscott:
Looking good! (To be honest I expected more of a back-and-forth design discussion rather than just simply accepting my proposal, but of course I am right, so this is easier haha)

Any idea why the CI is failing? Will hold off on reviewing till you've fixed that.

GeigerJ2:
> Looking good! (To be honest I expected more of a back-and-forth design discussion rather than just simply accepting my proposal, but of course I am right, so this is easier haha)

Hehe, I had the back-and-forth discussion locally with Claude. Will post the main points on the original issue. No need to do an in-depth review here for now, maybe we change the design, after all (implementation was just one prompt away, so I just went with it)

> Any idea why the CI is failing? Will hold off on reviewing till you've fixed that.

Was just a flaky test!

src/aiida_workgraph/tasks/builtins.py
```diff
@@ -107,14 +112,22 @@ class Map(Zone):
...
+
+    @property
+    def value(self) -> BaseSocket:
+        """Placeholder for the current entry's value while iterating the source."""
+        return self._map_item_task().outputs.value
+
+    @property
+    def key(self) -> BaseSocket:
+        """Placeholder for the current entry's key while iterating the source."""
+        return self._map_item_task().outputs.key
```
elinscott, line 127: Clearly here the changes you have made are superficial and the idea of the `item` persists internally. Is there any merit to a proper refactor that also changes things internally?

elinscott (approved):
Looks good! I also checked against my tests and the new behaviour worked as desired.

That said, this is a purely user-facing rename — z.value still just proxies to the map_item task's outputs, so the idea of an "item" persists internally. Maybe this design is ok, maybe not.

elinscott:
N.B.: the (sensible) naming choice of `.value` here makes #786 even more pressing, because currently `.value` can appear in two very different contexts. Once we squash the rogue cases where you need to explicitly reference `.value` for `TaggedValue` instances, we remove another potnetial point of confusion.

src/aiida_workgraph/tasks/builtins.py
GeigerJ2: Hehe, yes, true that! So, the main conclusion is that `Map` is fundamentally a _functional_ concept, so we could change things to a _functional_ API, as well. That means we would drop the `with` context-manager form that we otherwise recommend. Right now, it's nice bc it's consistent throughout, but we are somewhat forcing the functional concept of `Map` into the same shape as the control flow elements `If/While`, and that makes the usage of `Map` a bit weird. In this PR, I just focused on the user-facing API of accessing the elements, basically just removing the intermediate `item` namespace.

PR #795, opened 2026-06-24
🔧 `mypy`: enable strict type checking on `src`

GeigerJ2:
The mypy pre-commit hook excluded all of `src/`, so nothing under it was ever actually type-checked. This turns it on with a strict config and gets the already-clean modules checked, without a big annotation push up front. The main decisions:

- `strict = true` is enforced everywhere; modules that aren't annotated yet are parked in an `ignore_errors` list that shrinks one module at a time. Global strictness is never relaxed to silence errors, so the list only ever gets emptied (the opposite of ratcheting strictness up flag by flag).
- On top of `strict`, a few off-by-default error codes are enabled (`deprecated`, `possibly-undefined`, `mutable-override`, `truthy-iterable`, `exhaustive-match`, etc.); all pass on the checked modules.
- `yaml` and `tabulate` are checked against real stubs (`types-PyYAML`, `types-tabulate`, pinned to the runtime package versions); the genuinely untyped deps (`node_graph`, `aiida_pythonjob`, `kiwipy`) stay on `ignore_missing_imports`.
- The hook runs with `language: system` and `pass_filenames: false` against a fixed `src/aiida_workgraph` target, so a commit, `pre-commit run --all-files`, and a bare `mypy` all check the whole tree (an incremental commit can't pass while breaking a caller in an unchanged module). `mypy` is pinned `~=2.1.0` in the `pre-commit` extra, matching aiida-core.

`explicit-override` is left for a follow-up, since requiring `@override` means a sweep across the existing class hierarchies.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,6 +61,5 @@ repos:
       (?x)^(
         .github/.*|
         docs/.*|
-        tests/.*|
-        src/.*
+        tests/.*
       )$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,10 +53,10 @@ repos:
     name: mypy
     entry: mypy
     args: [--config-file=pyproject.toml, --pretty]
-    language: python
+    language: system
     types: [python]
     require_serial: true
-    pass_filenames: true
+    pass_filenames: false
     exclude: >-
       (?x)^(
         .github/.*|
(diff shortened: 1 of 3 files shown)


elinscott (approved):
> The mypy pre-commit hook excluded all of src/
... or, indeed, did not exist in the first place 🤦 

I would also be in favour of a github test that runs mypy without the list of files to skip, just so we don't forget to eventually get that list down to zero.

PR #797, opened 2026-07-02
🐛 Engine fixes for Map-zone cloning, gather races, and per-task (de)serializers

elinscott:
## Summary

Three engine-level fixes uncovered by porting a production workflow with per-orbital Map zones. Each failure mode below is independent; they ship together because the reproducing workflow unearths all three.

## 1. `@task` decorator: spec-level `deserializers` / `serializers`

`@task` now accepts `deserializers=` / `serializers=` kwargs and stores them on the spec. `PyFunctionTask.execute` falls back to the spec-level values when no per-call kwarg is provided, in both the sync and async execution branches.

Motivation: provides a mechanism to opt out of the default `StructureData -> Atoms` auto-deserializer

```python
@task(deserializers={"aiida.orm.StructureData": "my_plugin.utils.identity"})
def analyze(structure):
    ...
```

Only meaningful for plain-Python (PyFunction) tasks; `CalcJob` / `WorkChain` / `calcfunction` tasks are unaffected (AiiDA owns their deserialization).

## 2. `_patch_cloned_tasks`: materialise dynamic-namespace sockets on cloned tasks

Map-zone task cloning looked up link targets with a bare `to_node.inputs[scoped_name]`. A dotted path into a *dynamic* namespace (e.g. `pseudos.O`) does not exist on a fresh clone, so relinking raised `AttributeError` at submit time. Cloning now resolves the target through `Graph._resolve_or_create_input_socket`, which materialises the child socket on demand.

## 3. ~`update_map_task_state`: gather race and failure propagation~ DROPPED (see comment below)

~Two related failure modes when a Map zone's iterations finish close together:~

- ~Race: `are_childen_finished` can return `True` for the map zone before every per-prefix `gather_item` clone has recorded its result in `ctx._task_results`. The gather loop then `KeyError`s. The method now bails out cleanly while any clone is still in flight; the next `continue_workgraph` tick re-enters once results are populated.~
- ~Failure masking: a failed mapped iteration leaves its `gather_item` clone SKIPPED (`on_task_failed` skips downstream children), so a check keyed on FAILED alone never fires and the gather loop `KeyError`s on the empty result, burying the original calculation failure under an unrelated trace. Gather clones that end SKIPPED or FAILED now propagate FAILED to the map task, preserving the real cause. Covered by a new test, `test_map_zone_failed_iteration`: `KeyError` before the fix, map task ends FAILED after.~

## Compatibility

Fix 2 requires `Graph._resolve_or_create_input_socket`, which is not yet in a released node-graph (introduced alongside the dynamic-namespace work targeting node-graph main). Draft until that release; fixes 1 and 3 have no new dependencies and can be split out if preferred.

### DIFF
--- a/src/aiida_workgraph/decorator.py
+++ b/src/aiida_workgraph/decorator.py
@@ -21,6 +21,8 @@ def _spec_for(
     outputs: Optional[SocketSpec] = None,
     catalog: str = None,
     error_handlers: Optional[Dict[str, ErrorHandlerSpec]] = None,
+    deserializers: Optional[Dict[str, str]] = None,
+    serializers: Optional[Dict[str, str]] = None,
 ) -> TaskSpec:
     # AiiDA process classes
     if inspect.isclass(obj) and issubclass(obj, (CalcJob, WorkChain)):
@@ -46,6 +48,8 @@ def _spec_for(
             out_spec=outputs,
             error_handlers=error_handlers,
             catalog=catalog or 'Others',
+            deserializers=deserializers,
+            serializers=serializers,
         )
         return spec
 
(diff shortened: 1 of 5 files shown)


elinscott:
Dropped fix for the race condition from this PR as it is separately addressed by #776

PR #798, opened 2026-07-02
🐛 Coerce workgraph node attributes to JSON-safe values on save

elinscott:
## Problem

Node attributes must survive aiida-core's `clean_value`, but workgraph data can carry Python objects it rejects, so `save_workgraph_data` raised when the attributes were stored.

What actually reaches the attribute path with such values is error-handler `kwargs`, copied verbatim at two sites: a graph-level handler lands in `workgraph_error_handlers`, a task-level one inside `workgraph_data` under that task's spec. (An earlier version of this description claimed the trigger was an `enum.Enum` default in a task function signature — as measured in review, that raises earlier, in `general_serializer` during `to_engine_inputs()`, and never reaches the attribute path.)

## Change

`_ensure_json_safe` recursively coerces workgraph data before it is assigned to the process node:

- Enum members (values and mapping keys) are unwrapped to their `.value`
- the lossy `str()` fallback is gated on `clean_value` rather than `json.dumps`: only values `clean_value` genuinely rejects are stringified, while everything storage coerces itself (`set`/`frozenset` to list, numpy scalars to Python scalars, `BaseType` to its value) passes through untouched
- any `Mapping` has its keys coerced — `clean_value` does not inspect keys, so e.g. a tuple-keyed `MappingProxyType` otherwise passes the helper and fails in the database driver at `store()`
- one-shot iterators are materialized to lists — `clean_value` exhausts them as a side effect of validation, so an empty list would be stored otherwise

## Notes

- The `str()` fallback is lossy by design (e.g. arbitrary objects become their `str()`)
- The `.value` unwrap is limited to `enum.Enum`, so unrelated objects that merely expose a `.value` attribute are not silently unwrapped
- `NaN`/`inf` still fail at store: the `float` short-circuit precedes the gate; unchanged from `main`
- `bytes` pass through and store as a list of ints, matching what storage does without the helper

## Testing

Regression tests in `tests/test_utils.py` cover:

- the plain-Enum unwrap, key coercion, and the lossy fallback
- pass-through of `clean_value`-coercible values (`set`, `numpy.int64`, `orm.Int`) including an end-to-end store round-trip
- the non-`dict` `Mapping` and iterator cases above
- the `save_workgraph_data` wiring through `WorkGraph.save()` for both error-handler sites; these fail if the `_ensure_json_safe` call sites are removed

### DIFF
--- a/src/aiida_workgraph/utils/__init__.py
+++ b/src/aiida_workgraph/utils/__init__.py
@@ -18,6 +18,7 @@ from node_graph.socket import TaggedValue
 from node_graph.socket_spec import SocketSpec
 from aiida.orm.utils.serialize import serialize
 from aiida_workgraph.orm.utils import deserialize_safe
+import json
 from copy import deepcopy
 
 LOGGER = logging.getLogger(__name__)
@@ -268,6 +269,30 @@ def clean_pickled_task_executor(tdata: Dict[str, Any]) -> None:
             tdata['error_handlers'][name] = RuntimeExecutor.from_callable(UnavailableExecutor).to_dict()
 
 
+def _ensure_json_safe(value: Any) -> Any:
+    """Recursively ensure all values in a nested structure are JSON-serializable.
+
+    Node attributes must be JSON-serializable.  Workgraph data may contain
+    non-serializable Python objects (e.g. enum defaults from task function
+    signatures).  This function converts them to safe representations.
+    """
+    if value is None or isinstance(value, (bool, int, float, str)):
+        return value
+    if isinstance(value, dict):
+        return {k: _ensure_json_safe(v) for k, v in value.items()}
+    if isinstance(value, (list, tuple)):
+        return [_ensure_json_safe(v) for v in value]
+    try:
+        json.dumps(value)
+        return value
+    except (TypeError, ValueError):
+        pass
+    # Unwrap value-like objects (enums, etc.)
+    if hasattr(value, 'value') and not callable(value.value):
+        return _ensure_json_safe(value.value)
+    return str(value)
+
+
 def save_workgraph_data(node: Union[int, orm.Node], inputs: Dict[str, Any]) -> None:
     from aiida_workgraph.engine.workgraph import WorkGraphSpec
 
@@ -286,9 +311,9 @@ def save_workgraph_data(node: Union[int, orm.Node], inputs: Dict[str, Any]) -> N
     node.task_states = task_states
     node.task_processes = task_processes
     node.task_actions = task_actions
-    node.workgraph_data = wgdata
-    node.workgraph_data_short = short_wgdata
-    node.workgraph_error_handlers = wgdata.pop('error_handlers', {})
+    node.workgraph_data = _ensure_json_safe(wgdata)
+    node.workgraph_data_short = _ensure_json_safe(short_wgdata)
+    node.workgraph_error_handlers = _ensure_json_safe(wgdata.pop('error_handlers', {}))
     graph_inputs = dict(inputs.pop('graph_inputs', {}))
     tasks = dict(inputs.pop('tasks', {}))
     tasks['graph_inputs'] = graph_inputs


GeigerJ2:
Had a proper look at this, including where the data actually ends up. The approach is right and I'd keep the `str()` fallback. Two things.

**Gate the fallback on `clean_value` rather than `json.dumps`.** It's the more generous of the two: it turns any non-`str` iterable into a list, unwraps `BaseType`, coerces `numpy` integers. Currently, `_ensure_json_safe` runs immediately before it, so anything in that gap gets stringified before `clean_value` ever sees it. Measured through `WorkGraph.save()` with the value in error-handler `kwargs`: `{1, 2, 3}` stored as `[1, 2, 3]` on `main` and `'{1, 2, 3}'` here, `numpy.int64(7)` as `7` and now `'7'`, `orm.Int(3)` as `3` and now `'uuid: <uuid> (unstored) value: 3'`.

```python
try:
    clean_value(value)
except ValidationError:
    return str(value)
return value
```

Nothing is given up: everything `clean_value` genuinely rejects (`Enum`, `datetime`, `Path`, `UUID`, etc.) fails `json.dumps` too, so the fallback still catches all of it. Only the value path though, `_ensure_json_safe_key` should keep its whitelist, since `clean_value` doesn't look at dict keys at all and a tuple key otherwise reaches the DB and fails there.

<details>
<summary>MWE: a <code>set</code> in error-handler kwargs, and where <code>clean_value</code> sits</summary>

```python
from node_graph.error_handler import normalize_error_handlers

load_profile()


@task()
def add(x: int = 1, y: int = 2):
    return x + y


def handle(task, **kwargs):  # never runs, only stored
    return 'retrying'


wg = WorkGraph(
    'mwe',
    error_handlers=normalize_error_handlers(
        {'h': {'executor': handle, 'exit_codes': [1], 'kwargs': {'tags': {1, 2, 3}}}}
    ),
)
wg.add_task(add, name='add1')
wg.save()

stored = orm.load_node(wg.process.pk).base.attributes.get('workgraph_error_handlers')
print(repr(stored['h']['kwargs']['tags']))
```

```
main                    [1, 2, 3]      <- list
this PR                 '{1, 2, 3}'    <- str
gated on clean_value    [1, 2, 3]      <- list
```

`clean_value` is not being skipped, it still runs on every attribute. It just gets handed a string instead of the original value:

```
WorkGraph.save()
  aiida_workgraph/engine/workgraph.py   on_create()
  aiida_workgraph/utils/__init__.py     save_workgraph_data()   <- _ensure_json_safe(wgdata) runs here
  aiida_workgraph/orm/workgraph.py      setter()                <- node.workgraph_data = ...
  aiida/orm/nodes/attributes.py         NodeAttributes.set()
  aiida/storage/psql_dos/orm/nodes.py   SqlaNode.set_attribute()
                                          -> clean_value(value)
```

</details>

**Nothing exercises the `save_workgraph_data` path.** Every test calls `_ensure_json_safe` directly on a hand-built dict, so the helper is covered but its wiring isn't: deleting the call at all three sites there leaves `tests/test_utils.py` green (13 passed). Going through `WorkGraph.save()` also shows the trigger isn't quite the one in the description: an `Enum` default in a task signature raises earlier, in `general_serializer` during `to_engine_inputs()`. What does reach the attribute is error-handler `kwargs`, copied verbatim, at two sites: a graph-level handler lands in `workgraph_error_handlers`, a task-level one inside `workgraph_data` under that task's spec. Happy to push the tests for both I have locally.

elinscott:
Thanks @GeigerJ2  — adopted both points.

**Gate**: thanks, I wasn't aware of `clean_value`, that's much nicer. Now `{1, 2, 3}` stored as `[1, 2, 3]`, `numpy.int64(7)` as `7`, `orm.Int(3)` as `3`.

**Wiring**: added tests through `WorkGraph.save()` for both sites and verified they fail when the three `_ensure_json_safe` call sites are removed. (Edge case spotted by Claude: removing only the `workgraph_data_short` wrap would still pass, since no test puts a non-storable value where only the short-JSON path sees it. Don't think we need to worry about this.)

Adversarial testing also turned up two further holes, both now fixed:
- a non-`dict` `Mapping` (e.g. `MappingProxyType` with a tuple key) skipped the `dict` branch, passed `clean_value`, and failed at `store()` in the database driver — the mapping branch now tests `Mapping`
- a generator was exhausted by `clean_value` as a side effect of validation and stored as `[]` — one-shot iterators are now materialized to lists first

Updated the PR description accordingly.

GeigerJ2 (approved):
Great, thanks @elinscott!

PR #803, opened 2026-08-02
WorkGraph Process and DAG stepper

agoscinski:


src/aiida_workgraph/engine/process.py
```diff
@@ -0,0 +1,220 @@
...
+    from aiida_workgraph import WorkGraph
+
+__all__ = ('WorkGraphProcess', 'WorkGraphSpec')
+
+
+class WorkGraphSpec(WorkChainSpec):
+    WORKGRAPH_DATA_KEY = 'workgraph_data'
+
+
+class WorkGraphProcess(WorkChain):
```
agoscinski, line 33: The inherit by WorkGraph from WorkChain feels very awkward. They have different execution models and it seems now that you are extending them to match WorkGraph concepts (e.g. awaitable_barrier). I think one can get a cleaner code by using AiidaProcess class as starting point and some functionalities that you want to share between both you push into that class. Especially in terms of future changes, any change in WorkChain directly affects WorkGraph in this design, we don't want this.

PR #805, opened 2026-08-06
🐛 Read the kill message under the key plumpy sends

elinscott:
# Problem

`verdi process kill` on a WorkGraph fails and the graph keeps running:

```
$ verdi process kill 122058
Report: Request to kill Process<122058> sent.
Error: Failed to kill Process<122058>: module 'plumpy.process_comms' has no attribute 'MESSAGE_KEY'
```

`WorkGraphEngine.message_receive` reads the message text under `process_comms.MESSAGE_KEY` but  **plumpy has no such name** — it is `MESSAGE_TEXT_KEY`. That method is the process's RPC subscriber (`plumpy/processes.py:323` registers it against the pid), so...
- the `AttributeError` is raised while handling the incoming call
- kiwipy returns the exception to the caller as the result
- `Process.kill` is never reached
- and the graph carries on unharmed!

`verdi process pause` fails the same way.

The rename is not recent. It landed in `plumpy v0.24.0`, and this package requires `aiida-core>=2.7.3`, whose oldest supported release already pins `plumpy~=0.25.0`.

# Changes

- Read the text from `MESSAGE_TEXT_KEY`, the key `MessageBuilder.pause` and `MessageBuilder.kill` actually write.
- Pass it as `msg_text`. The rename travelled with a signature change — `Process.pause(msg_text=None)` and `Process.kill(msg_text=None, force_kill=False)` no longer accept `msg`. Fixing only the constant swaps the `AttributeError` for `TypeError: kill() got an unexpected keyword argument 'msg'`: still broken, and quieter.
- Forward `force_kill`, which the kill message also carries and which was being dropped, so `verdi process kill --force` was silently downgraded to a normal kill. The defaults come from `MessageBuilder.kill()`, mirroring `plumpy.Process.message_receive`, so a message built by an older client without the key still means a plain kill.
- `verdi process kill --force` now takes effect on WorkGraphs; as for any AiiDA process it skips the child-kill loop, so running `CalcJob`s are left on the scheduler.

# Testing

`test_message_receive_reads_plumpy_message_text` drives `message_receive` with a message built by plumpy's own `MessageBuilder`, against a stub whose `pause`/`kill` carry the real signatures, so no broker or daemon is needed. Three mutants, each failing:

| mutant | failure |
|---|---|
| the code as it is today | `AttributeError: module 'plumpy.process_comms' has no attribute 'MESSAGE_KEY'` |
| constant renamed, `msg=` kept | `TypeError: kill() got an unexpected keyword argument 'msg'` |
| `force_kill` forwarding removed | `assert ('forced', False) == ('forced', True)` |

The full send path was also exercised through the real `RemoteProcessThreadController` and `MessageBuilder`, with a JSON round-trip standing in for the broker:

```
verdi process kill          → ('kill', 'Killed through `verdi process kill`', False)
verdi process kill --force  → ('kill', 'Force killed through `verdi process kill`', True)
message without force_kill  → ('kill', 'old client', False)
```

Everything above is reproduced. That `Process.kill` then terminates a live graph is not asserted here — it needs a broker, and this change only ensures the call is reached with the right arguments.

Note for whoever tests this against a running graph: with `force_kill=True` plumpy transitions straight to `KILLED`, so `aiida-core`'s child-kill loop is skipped and running `CalcJob`s are orphaned. That is aiida-core's `-F` semantics for any process, not a change made here — but before this fix `-F` did nothing at all, so it will be newly visible.

# Update
The engine no longer carries a copy of plumpy's `message_receive`: it handles its own custom task actions and delegates every plumpy intent to the base implementation.


elinscott:
@edan-bainglass @GeigerJ2

GeigerJ2 (approved):
LGTM, thanks @elinscott!

